PR #25, opened 2025-01-21
Increase session_id length specifier in table online to 32 chars

nielsrune:
Increase session_id length specifier in table online to 32 chars

PHP default session id length is 32 chars.
With only 30 chars in db, this is crashing the app on login and throwing the following log:

_PHP Warning:  pg_query(): Query failed: ERROR:  value too long for type character varying(30) in .../includes/db_query.php on line 121, referer: http://.../index/index.php_

See also this regarding deprecation as of PHP 8.4: https://www.php.net/manual/en/migration84.deprecated.php#migration84.deprecated.session

### DIFF
--- a/includes/betweenUpdates.php
+++ b/includes/betweenUpdates.php
@@ -366,10 +366,6 @@ $qtxt="SELECT column_name FROM information_schema.columns WHERE table_name='regn
 if (!$r=db_fetch_array(db_select($qtxt,__FILE__ . " linje " . __LINE__))) {
 	db_modify("ALTER table regnskab ADD column sms integer",__FILE__ . " linje " . __LINE__);
 }
-$qtxt="SELECT column_name FROM information_schema.columns WHERE table_name='online' and column_name='session_id'";
-if (!$r=db_fetch_array(db_select($qtxt,__FILE__ . " linje " . __LINE__))) {
-	db_modify("ALTER TABLE online ALTER COLUMN session_id TYPE varchar(32)",__FILE__ . " linje " . __LINE__);
-}
 include ('../includes/online.php');
 
 ?>
--- a/includes/betweenUpdates.php
+++ b/includes/betweenUpdates.php
@@ -366,6 +366,10 @@ $qtxt="SELECT column_name FROM information_schema.columns WHERE table_name='regn
 if (!$r=db_fetch_array(db_select($qtxt,__FILE__ . " linje " . __LINE__))) {
 	db_modify("ALTER table regnskab ADD column sms integer",__FILE__ . " linje " . __LINE__);
 }
+$qtxt="SELECT column_name FROM information_schema.columns WHERE table_name='online' and column_name='session_id'";
+if (!$r=db_fetch_array(db_select($qtxt,__FILE__ . " linje " . __LINE__))) {
+	db_modify("ALTER TABLE online ALTER COLUMN session_id TYPE varchar(32)",__FILE__ . " linje " . __LINE__);
+}
 include ('../includes/online.php');
 
 ?>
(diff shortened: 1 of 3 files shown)


index/install.php
```diff
@@ -216,7 +217,7 @@
 	$qtxt = "INSERT INTO regnskab (regnskab, dbhost, dbuser, db, version,bilag) values ";
 	$qtxt.= "('$db_navn' ,'$host', '$db_bruger', '$db_navn', '$version','0')";
 	db_modify($qtxt,__FILE__ . " linje " . __LINE__);
-	$qtxt = "CREATE TABLE online (session_id varchar(30), brugernavn text, db varchar(30), dbuser varchar(30), rettigheder varchar(30), ";
+	$qtxt = "CREATE TABLE online (session_id varchar(32), brugernavn text, db varchar(30), dbuser varchar(30), rettigheder varchar(30), ";
```
logicguy1, line 220: This is good, however we need to also update the inbetween updates file such that all exsisting systems of saldi has this fix applied as well

Checkout [This code](https://github.com/DANOSOFT/saldi/blob/master/includes/betweenUpdates.php#L363-L369) where a similar fix was applied

logicguy1 (changes requested):

index/install.php
nielsrune: Thank for the link!
I have updated this PR.
However, this is not sufficient for existing installations as `betweenUpdates.php` are being included to late in `login.php`

The bug this PR adresses happens here at line 298:
https://github.com/DANOSOFT/saldi/blob/377e1dc8733ed9ec06c7fc96123890f3681a831d/index/login.php#L298

whereas `betweenUpdates.php` are not included until line 488:
https://github.com/DANOSOFT/saldi/blob/377e1dc8733ed9ec06c7fc96123890f3681a831d/index/login.php#L488

Any preferences for a way forward?
My best suggestion is to run a single "dirty in-between alteration" much earlier during login.

logicguy1: Ah ofc. I think the best way forward would be to just modify login.php to include a lookup to the schema table to see if the column already has been updated, then update the column with the correct size.

We do this occasionally to help with migration in production, but we usually move them into between updates and later their concrete update file such as [this one](https://github.com/DANOSOFT/saldi/blob/master/includes/opdat_4.0.php)

So it is just important to add some notes as to why betweenupdates was not used, so it does not get "cleaned up" later in our development process

includes/betweenUpdates.php
```diff
@@ -366,6 +366,10 @@ function ensureTableAndColumns($db, $tableName, $expectedColumns, $renameColumns
 if (!$r=db_fetch_array(db_select($qtxt,__FILE__ . " linje " . __LINE__))) {
 	db_modify("ALTER table regnskab ADD column sms integer",__FILE__ . " linje " . __LINE__);
 }
+$qtxt="SELECT column_name FROM information_schema.columns WHERE table_name='online' and column_name='session_id'";
+if (!$r=db_fetch_array(db_select($qtxt,__FILE__ . " linje " . __LINE__))) {
```
logicguy1, line 370: Note that this checks if the column does not exsist, something like
```
SELECT column_name, data_type FROM information_schema.columns WHERE table_name = 'online' AND column_name = 'session_id'
```
Would also fetch the datatype

nielsrune: Better check now in `login.php`

index/install.php
nielsrune: Done

logicguy1:
Looks perfect niels, feel free to also check the vareliste later, it will have had a massive overhaul.

I will merge this now.

logicguy1:
Oops

PR #26, opened 2025-01-21
Fix several warnings thrown on admin login

nielsrune:
This fixes several issues found during first admin login after a fresh install in a docker setup.
See commit messages for further info.

Special attention to L62-L72 in index/index.php which has been commented out temporarily.  If needed, the block should be placed in another location AFTER headers (setcookie...)


index/login.php
```diff
@@ -77,13 +77,16 @@
...
 
-print "<!--";
-$timezone = system("timedatectl | grep \"Time zone\"");
-print "-->";
-list($tmp,$timezone) = explode(":",$timezone);
-list($timezone,$tmp) = explode("(",$timezone);
-$timezone = trim($timezone);
-if (!$timezone) $timezone = 'Europe/Copenhagen';
+#print "<!--";
+$timezone = system("timedatectl 2>/dev/null | grep \"Time zone\"", $errcode);
```
logicguy1, line 81: $errorcode not defined before usage in system function

includes/db_query.php
```diff
@@ -142,7 +142,7 @@ function db_modify($qtext, $spor) {
 		}
 #20190704 END
 		
-		$db=trim($db);
+		$db=trim($db ?? '');
```
logicguy1, line 145: PHP code not compatible for PHP version 5, use std function if_isset($var, $default) instead, as this will ensure cross comparability with all versions of PHP equal to or later than 5.

logicguy1 (changes requested):

index/login.php
nielsrune: Yes, `$errcode` is assigned by reference internally. Therefore initialization is not needed.

Try running this code in a terminal:
```
<?php
$good_command = system("ls >/dev/null", $errcode1);
var_dump($errcode1);  // -> int(0) as the ls command was succesful

$bad_command = system("invalid_command >/dev/null", $errcode2);
var_dump($errcode2); // -> int(127) as the command invalid_command was not found
?>
```

includes/db_query.php
nielsrune: Ok, trivial here, but why are you requiring comparability with PHP 5.0, which was [end of life](https://www.php.net/eol.php) nearly 20 years ago??
Even the last point release 5.6 was end of life 6 years ago, and as of now even PHP 8.0 is end of life.

Surely you are not running PHP 5.x on production(?!)
If so, you should really prioritize migrating fully to a more recent version.
Maybe not bleeding edge, but at least aim for [8.2 which is getting security fixes until the end of 2026](https://www.php.net/supported-versions.php).

This will also make possible major code revision, for instance a much more clean and secure uniform database handling (ask in in doubt).

By the way, did a search, and you already have the null coalescing operator two other places:
https://github.com/DANOSOFT/saldi/blob/377e1dc8733ed9ec06c7fc96123890f3681a831d/index/dashboardIncludes/revenue_graph.php#L48
https://github.com/DANOSOFT/saldi/blob/377e1dc8733ed9ec06c7fc96123890f3681a831d/index/dashboardIncludes/revenue_graph.php#L63

Finally, you are already using the ternary operation several places which was a new feature in 5.3

logicguy1: I believe we still have 2 out of 7 production environments that are still yet to be upgraded from php 5.5 to 8.3.

Why this has not been done earlier i do not know, but we are running into a lot of migration errors, as 8.x introduced a lot of breaking changes. 

So we are in progress on the migrations, ensuring everything goes as smoothly as it can with hundreds of users on each server.

logicguy1: Yep, sorry, i should get this updated in the development guidelines, it's something that bites you in the ass sometimes as a developer, so our entire team is hopeful the migration is done sooner than later.

The code comment in index/index.php is important as it checks that the login screen is never shown next to the sidebar, as that is horrible UX. As it is a direct redirect it might be better to just throw in an exit; or die(); to exit out of the script and let it redirect the entire page.

nielsrune: Ok, I must admit I don't fully understand the flow - the code is really complex and I'm only used to the classic design.

But just be aware, for PR #28 to work (husk mig), no output can be sent before headers, which includes `setcookie()`.

So either the javascript code must be moved further down after the 3 setcookie calls (last one in line 99), or the check must somehow be done on the server side and in case a redirect is needed that can be done using `header(Location)`.

logicguy1: Yeah, the login process is one we just don't touch often for that reason, the redirect gets sent such that you arent able to login if it is not on the top aka not in an iframe. So the user wont be able to check the "Husk mig" checkbox because they will be redirected before that.

When we get all the codebase in one central state using PHP 8.2 the plan is to start writing some proper development documentation, mapping out some of these more complex processes.

nielsrune: Okay hold your horses.
Just did a test in incognito tab and it seems to work even with the js being sent on top before `<html>`.
right now I can't figure out why it's actually working, but if it can be confirmed that huskmig is in deed working after apply this PR not further action is needed.
(In time confirmation should also include the PR #28  regarding the language setting)

PR #28, opened 2025-01-25
This fix broken language selection in admin -> settings.

nielsrune:
This fix broken language selection in admin -> settings.

Together with new function get_root() it is ensure, that the cookie 'languageId' is set across the entire app folder, not individual subfolders, i.e. index/ and admin/

Also minor correction and highlighted questions.


admin/admin_settings.php
```diff
@@ -119,11 +123,14 @@
 
 $q = db_select("select * from brugere where brugernavn = '$brugernavn'",__FILE__ . " linje " . __LINE__);
 $r = db_fetch_array($q);
-if ($brugerId=$r['id']) {
+/** The following test is always false as this is the only occurence of $brugerId in the entire codebase.
```
logicguy1, line 126: The variable bruger_id exsists, however the variables that get created don't seem to be called anywhere

admin/admin_settings.php
```diff
@@ -100,10 +101,13 @@
 	}
 	update_settings_value("nyhed", "dashboard", $newssnippet, "The news snippet showen to all admin accounts on this system");
 	db_modify($qtxt,__FILE__ . " linje " . __LINE__);
-	$qtxt="update settings set var_value='$languageId' where var_name='languageId'";
+	$qtxt="update settings set var_value='$languageId' where var_name='language_id'";
```
logicguy1, line 104: What is the logic behind changing languageId to language_id?

admin/admin_settings.php
nielsrune: Actually, I have misread the if statement as a comparison. But now I see it is an assignment to `$brugerId`.

In this case, a more readable code would be just `isset($r['id']` as it is doing nothing more. And `$brugerId` is not used anywhere else

admin/admin_settings.php
nielsrune: The logic being that `language_id` is the value set in the table field `var_name` :-)
See here:
https://github.com/DANOSOFT/saldi/blob/377e1dc8733ed9ec06c7fc96123890f3681a831d/index/install.php#L272-L273

logicguy1:
I just need to run this though Peter before we can pull the PR

nielsrune:
This shouldn't be merged before #26, see comment just now.
I don't recall why the javascript code in index.php was not handled in this one :-(

Issue #31, opened 2025-01-28
Question regarding if_isset


nielsrune:
Just a question, but with the function's name in mind shouldn't the function test for `isset()` and not just `if()`? (line 213).

Also why is `$var` passed in by reference `&$var` when the function does not change the variable? (line 201)

https://github.com/DANOSOFT/saldi/blob/377e1dc8733ed9ec06c7fc96123890f3681a831d/includes/std_func.php#L201-L216

nielsrune:
To elaborate on the somewhat misleading function name, see the the 11th row in the first table here https://www.php.net/manual/en/types.comparisons.php :
`$x = 0;`

A true `isset()` should return true because `$x` is indeed given a value of zero. What the custom made function `if_isset` is doing is a boolean comparison which in the case would return false, because zero is considered false.

Maybe somewhere the code depends on this behavior, so a switch to `isset` will introduce bugs in other places.
But I'm currently debugging the `admin/restore.php` where the `if_isset` cannot be used despite the name, because the variable being tested is actually set to zero - which is good since it's an errorcode.

A dummy code to show what I mean:
```<? php
// set a variable to zero
$array['a']['b'] = 0;

// this will print "isset"
if (isset($array['a']['b'])) echo "isset";

// this is the current behavior of your if_isset
// and will not print anything
if ($array['a']['b']) echo "isset";
?>
```

logicguy1:
I raised this to the team and we believe it is best to modify the function to use isset, and then just manage any edge cases after. I don't think any features in saldi require this functionality, at least from talking to the other developers.

Loe4g:
I believe this issue has been addressed in this commit; [47c8e1f13b5](https://github.com/DANOSOFT/saldi/commit/47c8e1f13b52a46ce63e36d77fbeeb1efa78f5b3). Please let us know within the next two weeks if any new observations arise. Also, if it has covered the bulk of those mentioned, then, like @logicguy1 said, any edge cases not covered can likely be handled using native PHP functions. 👍"

PR #36, opened 2025-02-01
Add hostname to psql

nielsrune:
Follow up to #24

`psql` hostname must be set when running in docker or otherwise separate hosts for Restore to work

### DIFF
--- a/admin/restore.php
+++ b/admin/restore.php
@@ -30,6 +30,7 @@
 // 20241108 PHR PHP8
 // 20250130 migrate utf8_en-/decode() to mb_convert_encoding
 // 20222706 MSC - Implementing new design
+// 20250201 Add hostname to psql
 
 @session_start();
 $s_id=session_id();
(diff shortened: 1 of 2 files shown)


admin/restore.php
```diff
@@ -278,7 +279,7 @@ function restore($filnavn,$backup_encode,$backup_dbtype){
   print "<!-- Saldi-kommentar for at skjule uddata til siden \n"; # Indsat da svar fra pg_dump kan resultere i besked genereres
```
nielsrune, line 279: Just a note - this is found several places.
A much better solution is to redirect stdout to /dev/null - and stderr to errorlog

admin/restore.php
```diff
@@ -111,15 +114,20 @@ function confirmSubmit()
...
+	}
 	if ($fejl) {
 		print "<script>javascript:document.body.style.cursor = 'default';</script>";
 		switch ($fejl) {
 			case 1: print "<BODY onLoad=\"javascript:alert('Filen er for stor - Kontroller upload_max_filesize i php.ini')\">";
 			case 2: print "<BODY onLoad=\"javascript:alert('Filen er for stor - er det en SALDI-sikkerhedskopi?')\">";
 		}
 	}
-	if (basename($_FILES['uploadedfile']['name'])) {
+	if (isset($_FILES['uploadedfile']['name']) && basename($_FILES['uploadedfile']['name'])) {
```
nielsrune, line 130: Is the basename check needed?
Like #31 this is a boolean comparison on a [string function](https://www.php.net/manual/en/function.basename.php) which can lead to [unexpected behavior](https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting)

nielsrune:
PR is updated after debugging some PHP warnings in the errorlog

nielsrune:
I really don't understand this commit done today: 6de73083db543d83178cc88e4768ddcb4c938d91

Just by looking at line 7 and 30, it seems that work was done on an out of sync local clone. 
It reverts this commit from December 2024 - 979da5e48a2668b566e4983e7126b450ec342109 - with no info regarding this in the commit message.

The commit message says *"Fixed error where it would not restore an account"* while the note in the beginning of the file says *"Implementing new design"*
And the date note and version specifier is bump a good way back in the past?

By the way and for what it's worth, the restore procedure itself seems to work fine at me without this commit, but with the proposed changes in this PR.

admin/restore.php
logicguy1: I think your observation is correct, we need to make sure the file exsists and has a valid file, maybe we should use an if(!empty()) on it.

admin/restore.php
```diff
@@ -292,7 +299,7 @@ function restore($filnavn,$backup_encode,$backup_dbtype){
 	if ($popup) {
 		print "<BODY ONLOAD=\"JavaScript:opener.location.reload();\"";
 		print "<meta http-equiv=\"refresh\" content=\"0;URL=../includes/luk.php\">";
```
nielsrune, line 301: Noticed this during rebasing. I'm sure you are aware, but it is extremely bad coding to put multiple body-tags inside a document. Also meta-tags are only meant to be inside the head-tags.
Sometime, you should clean up this mess across the entire codebase, since browsers may someday stop rendering this correctly without notice.

PR #49, opened 2025-02-12
Fix passing of NULL to explode during soring logic

nielsrune:
On the initial entry into `finans/kladdeliste.php` the following warning is thrown:
```
PHP Deprecated:  explode(): Passing null to parameter #2 ($string) of type string is deprecated in /var/www/html/danosoft/finans/kladdeliste.php on line 54, referer: http://localhost/danosoft/index/menu.php
PHP Stack trace:, referer: http://localhost/danosoft/index/menu.php
PHP   1. {main}() /var/www/html/danosoft/finans/kladdeliste.php:0, referer: http://localhost/danosoft/index/menu.php
PHP   2. explode($separator = ';', $string = NULL) /var/www/html/danosoft/finans/kladdeliste.php:54, referer: http://localhost/danosoft/index/menu.php
```

This is because, if the cookie `saldi_kladdeliste` does not exists, at line 54, `NULL` will be passed to `explode()`.

This PR rearranges the logic to make sure, no undefined variables are relied on. Also see specific line notes.

### DIFF
--- a/finans/kladdeliste.php
+++ b/finans/kladdeliste.php
@@ -41,21 +41,23 @@ $apiKey = db_fetch_array($query)["var_value"];
 include("../includes/online.php");
 include("../includes/topline_settings.php");
 
-if (!isset ($_COOKIE['saldi_kladdeliste'])) $_COOKIE['saldi_kladdeliste'] = NULL;
-
-$sort=isset($_GET['sort'])? $_GET['sort']:Null;
-$rf=isset($_GET['rf'])? $_GET['rf']:Null;
-$vis=isset($_GET['vis'])? $_GET['vis']:Null;
-print "<meta http-equiv=\"refresh\" content=\"150;URL=kladdeliste.php?sort=$sort&rf=$rf&vis=$vis\">";
-
-if (isset($_GET['sort'])) {
+list($sort,$rf,$vis) = [
+	isset($_GET['sort']) ? $_GET['sort'] : NULL,
+	isset($_GET['rf']) ? $_GET['rf']: NULL,
+	isset($_GET['vis']) ? $_GET['vis']: NULL
+];
+if (isset($sort)) {
 	$cookievalue="$sort;$rf;$vis";
 	setcookie("saldi_kladdeliste", $cookievalue, strtotime('+30 days'));
-} else list ($sort,$rf,$vis) = array_pad(explode(";", $_COOKIE['saldi_kladdeliste']), 3, null);
-if (!$sort) {
+} elseif (isset($_COOKIE['saldi_kladdeliste'])) {
+	list($sort,$rf,$vis) = array_pad(explode(";", $_COOKIE['saldi_kladdeliste']), 3, NULL);
+}
+else {
 	$sort = "id";
 	$rf = "desc";
 }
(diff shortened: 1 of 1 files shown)


finans/kladdeliste.php
```diff
@@ -41,21 +41,23 @@
 include("../includes/online.php");
 include("../includes/topline_settings.php");
 
-if (!isset ($_COOKIE['saldi_kladdeliste'])) $_COOKIE['saldi_kladdeliste'] = NULL;
```
nielsrune, line 44: This, to me, made no sense at all

finans/kladdeliste.php
```diff
@@ -41,21 +41,23 @@
...
-$rf=isset($_GET['rf'])? $_GET['rf']:Null;
-$vis=isset($_GET['vis'])? $_GET['vis']:Null;
-print "<meta http-equiv=\"refresh\" content=\"150;URL=kladdeliste.php?sort=$sort&rf=$rf&vis=$vis\">";
-
-if (isset($_GET['sort'])) {
+list($sort,$rf,$vis) = [
+	isset($_GET['sort']) ? $_GET['sort'] : NULL,
+	isset($_GET['rf']) ? $_GET['rf']: NULL,
+	isset($_GET['vis']) ? $_GET['vis']: NULL
+];
```
nielsrune, line 48: Initialize `$sort`, `$rf` and `$vis` with priority to `$_GET`

finans/kladdeliste.php
```diff
@@ -41,21 +41,23 @@
...
+if (isset($sort)) {
 	$cookievalue="$sort;$rf;$vis";
 	setcookie("saldi_kladdeliste", $cookievalue, strtotime('+30 days'));
```
nielsrune, line 51: Set or update cookie if `$sort` is not `NULL`

finans/kladdeliste.php
```diff
@@ -41,21 +41,23 @@
...
-} else list ($sort,$rf,$vis) = array_pad(explode(";", $_COOKIE['saldi_kladdeliste']), 3, null);
-if (!$sort) {
+} elseif (isset($_COOKIE['saldi_kladdeliste'])) {
+	list($sort,$rf,$vis) = array_pad(explode(";", $_COOKIE['saldi_kladdeliste']), 3, NULL);
+}
```
nielsrune, line 54: Reassign `$sort`, `$rf` and `$vis` if cookie exists

finans/kladdeliste.php
```diff
@@ -41,21 +41,23 @@
...
+else {
 	$sort = "id";
 	$rf = "desc";
 }
```
nielsrune, line 58: Set default values for first run.

finans/kladdeliste.php
```diff
@@ -41,21 +41,23 @@
...
+print "<meta http-equiv=\"refresh\" content=\"150;URL=kladdeliste.php?sort=$sort&rf=$rf&vis=$vis\">";
```
nielsrune, line 59: I don't know, why you are updating the page every 2½ seconds.
A html meta tag should no go inside a body tag as it does here.

logicguy1: I think this had something to do with always keeping the kladdeliste up to date if you were working with other people

PR #53, opened 2025-02-13
Fix multiple warnings in kassekladde

nielsrune:
Fix multiple warnings when passing NULL as a string argument etc.

[Technical background](https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation)

See specific line notes

### DIFF
--- a/finans/kassekladde.php
+++ b/finans/kassekladde.php
@@ -1,4 +1,3 @@
-
 <?php
 //                ___   _   _   ___  _     ___  _ _
 //               / __| / \ | | |   \| |   |   \| / /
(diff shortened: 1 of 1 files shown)


finans/kassekladde.php
```diff
@@ -1949,7 +1948,7 @@ function saveScrollPosition() {
 	
 	}
 	#############################################################################################################################
-	function kontroller($id, $bilag, $dato, $beskrivelse, $d_type, $debet, $k_type, $kredit, $faktura, $belob, $momsfri, $kontonr, $kladde_id, $afd, $projekt, $ansat, $valuta, $forfaldsdato, $betal_id, $lobenr) {
+	function kontroller($id, $bilag, $dato, $beskrivelse, $d_type, $debet, $k_type, $kredit, $faktura, $belob, $momsfri, $kladde_id, $afd, $projekt, $ansat, $valuta, $forfaldsdato, $betal_id, $lobenr) {
```
nielsrune, line 1951: `$kontonr` is never used in the function and all calls to the function passes NULL to this variable. Therefore removed

finans/kassekladde.php
```diff
@@ -294,7 +293,7 @@ function bogfor() {
 				$qtxt .= "betal_id='$betal_id[$x]' where lobenr='$x' and kladde_id='$kladde_id'";
 			}
 			db_modify($qtxt, __FILE__ . " linje " . __LINE__);
-			kontroller($id[$x], $bilag[$x], $dato[$x], $beskrivelse[$x], $d_type[$x], $debet[$x], $k_type[$x], $kredit[$x], $faktura[$x], $belob[$x], $momsfri[$x], $kontonr[$x], $kladde_id, $afd[$x], $projekt[$x], $ansat[$x], $valuta[$x], $forfaldsdato[$x], $betal_id[$x], $x);
+			kontroller($id[$x], $bilag[$x], $dato[$x], $beskrivelse[$x], $d_type[$x], $debet[$x], $k_type[$x], $kredit[$x], $faktura[$x], $belob[$x], $momsfri[$x], $kladde_id, $afd[$x], $projekt[$x], $ansat[$x], $valuta[$x], $forfaldsdato[$x], $betal_id[$x], $x);
```
nielsrune, line 296: `$kontonr` is always NULL, since `$x` is `0` and the array is not consecutively zero-indexed. Also removed from function definition as the argument was never used anyway

finans/kassekladde.php
```diff
@@ -1627,7 +1626,7 @@ function saveScrollPosition() {
 		} elseif ($kontrolkonto && $kontrolsaldo) {
 			print "<td align=right>" . dkdecimal($kontrolsaldo, 2) . "</td>\n";
 		}
-		if ($kontrolkonto && abs($saldo[$y]) > 0) {
+		if ($kontrolkonto && abs((float)$saldo[$y]) > 0) {
```
nielsrune, line 1629: `$saldo` is most often filled with NULL values, which cannot be passed to `abs()` as of PHP 8.1. Therefore casted to float which removes the warning

PR #79, opened 2025-06-13
Fix/discount only for saet zero

sawaneh1:
Show discount and gross profit only for order lines where 'saet' is NULL or 0, excluding discounts from set sales.

### DIFF
--- a/debitor/kassespor.php
+++ b/debitor/kassespor.php
@@ -521,45 +521,36 @@ $total_gross_profit = 0;
 		$dkksum[$x]=dkdecimal($sum[$x]+$moms[$x],2);
 	
 
-					$q_dg = db_fetch_array(db_select("
-					SELECT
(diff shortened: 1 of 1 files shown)


LuiHedlund02:
Closing this (June-2025 vintage). It targets an old `debitor/kassespor.php`
that has since been substantially rewritten — the corresponding query paths
are now around :609-638 and :732-783, and displayed discount is now derived
from the configured discount item (:759-776), which conflicts with this
patch's `$discount_val` approach. It also adds
`AND NOT (pris = 0 AND kostpris > 0)`, which is outside the set-line request
and would drop legitimate free/zero-price lines, overstating contribution
margin.

If the set-line BA policy is still wanted, let's open a fresh narrow PR
against the current report that preserves the discount-item presentation and
tests the zero-price and credit-note cases explicitly.

PR #171, opened 2026-01-23
Hide internal stock sync lines from invoices

logicguy1:
Order lines starting with "xc" in the description and having a zero price are now hidden from PDF invoices
These lines are used for internal stock synchronization and should not appear on customer-facing documents

When on a bigger project there may be multiple stock items that are used but not invoiced to the customer like network cables, external consultants etc. This is required to give the invoice a correct cost price.

The system is setup to check that the item is prefixed with xc and has a sales price of 0, this should avoid any accidental items getting hit by this filter, the xc keys were chosen because they are very close to each other on the keyboard

### DIFF
--- a/includes/formfunk.php
+++ b/includes/formfunk.php
@@ -90,6 +90,7 @@
 // 20240918 PBLM Added function betalingslink
 // 20241002 PHR 'Kontant' in texts replaced by text ID 370
 // 20250630 PHR Somebody has removed brackets in line 2285 & 2289 - Why !!!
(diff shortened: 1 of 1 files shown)


LuiHedlund02:
Closing this implementation. Internal stock-sync visibility should use an explicit order-line flag with defined form/tenant scope, rather than interpreting an `xc` description prefix. The current condition can suppress ordinary XC60/XC90 descriptions whenever their line total is zero, and applies across document types. Preserve that product text as text; a future scoped line-flag change needs dedicated print tests.

PR #210, opened 2026-02-20
Added stykliste liste to easily give insight into what items are "kits"

logicguy1:
image
(removed our price / cost price, this would normally be available to the end user)

Added a stykliste list that shows all the samlevarer that are created in the system, this is useful to quickly find a "Kit item"

Stock values are not included into this list as that would be quite the job to recursively fetch stock values of child items (as a stykliste can include other styklister), this would be great UX for end user however.

The icon i used is just a simple crate icon, i couldn't really think of another way to iconize a stykliste (perhaps a collection of boxes), but if anyone has any good ideas ill gladly update the PR

### DIFF
--- a/lager/lister/styklister.php
+++ b/lager/lister/styklister.php
@@ -1,0 +1,371 @@
+<?php
+//                ___   _   _   ___  _     ___  _ _
+//               / __| / \ | | |   \| |   |   \| / /
+//               \__ \/ _ \| |_| |) | | _ | |) |  <
+//               |___/_/ \_|___|___/|_||_||___/|_\_\
+//
+// ---- lager/lister/styklister.php --- lap 5.0.0 --- 2026.02.20 ---
+// LICENSE
+//
+// This program is free software. You can redistribute it and / or
+// modify it under the terms of the GNU General Public License (GPL)
+// which is published by The Free Software Foundation; either in version 2
+// of this license or later version of your choice.
+// However, respect the following:
+//
+// It is forbidden to use this program in competition with Saldi.DK ApS
+// or other proprietor of the program without prior written agreement.
+//
+// The program is published with the hope that it will be beneficial,
+// but WITHOUT ANY KIND OF CLAIM OR WARRANTY. See
+// GNU General Public License for more details.
+//
+// Copyright (c) 2026 saldi.dk aps
+// ----------------------------------------------------------------------
+// 20260220 LOE  - Created: Vareliste view filtered to samlevare (stykliste) items.
+
+@session_start();
+$s_id = session_id();
+
+$css = "../../css/standard.css?v=20";
+
+include("../../includes/std_func.php");
+include("../../includes/connect.php");
+include("../../includes/online.php");
+include("../../includes/stdFunc/dkDecimal.php");
+
+$valg = "Styklister";
+$returside = if_isset($_GET, get_relative()."index/menu.php", "returside");
+include("topLineVarer.php");
+
+include(get_relative() . "includes/grid.php");
+
+$vatOnItemCard = get_settings_value("vatOnItemCard", "items", "on") == "on"
+    ? true : false;
+
+// Columnconfig
+$columns = array();
+
+$columns[] = array(
+    "field" => "varenr",
+    "headerName" => "Vare Nr.",
+    "render" => function ($value, $row, $column) {
+        $url = "../../lager/varekort.php?id=$row[id]&returside=lister/styklister.php";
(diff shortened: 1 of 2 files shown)


LuiHedlund02:
@logicguy1: still current? Please rebase-check and: run `$returside` through `nav_sanitize_returside()`, reuse the existing VAT-inclusive price derivation instead of a new join, and put the column headers through `findtekst()`.
Lead reviewer: @ZaynSaul.

LuiHedlund02:
@logicguy1 please make the list ready before @ZaynSaul’s review: sanitize returside via the shared helper, avoid another independent VAT-price calculation and route new labels through translations. Verify return navigation and VAT-inclusive/exclusive display against the existing item list.

ZaynSaul (changes requested):
Verified Lui's three points against the current diff (still just the original Feb 20 commit, nothing pushed since his request):

**1. `$returside` sanitization - confirmed real, should be fixed.** `styklister.php` takes `$_GET['returside']` raw (`if_isset($_GET, get_relative()."index/menu.php", "returside")`) with no validation, then interpolates it straight into `href` attributes in `topLineVarer.php`'s nav buttons. `includes/stdFunc/navStack.php`'s `nav_sanitize_returside()` exists specifically to guard this exact pattern (blocks attribute-breaking characters, HTML entity smuggling, off-site targets) and is already used in 8 other files including `lager/varekort.php` - the very page this new file links to for item details. This new file should route through it too.

**2. VAT-price duplication - real, but not a live bug today.** The `momspris` CASE expression here is a byte-for-byte copy of `lager/lister/vareliste.php`'s own `momspris` column (same box7/box4/sm join chain). It's correct right now, but it's an independent copy that won't stay in sync if the formula changes in one place and not the other - which is a demonstrated failure mode in this codebase (a couple of PRs back a stock-reorder formula diverged exactly this way between two pages). Worth extracting to a shared helper while this is fresh, but I wouldn't block merge on this alone if the first point is fixed.

**3. Hardcoded Danish column headers - pushing back a bit.** Checked `vareliste.php`, which this file is directly modeled on (same columns, same query shape): it also hardcodes Danish `headerName` strings ("Vare Nr.", "Navn", "Varemærke", etc.), not `findtekst()`. So this isn't a regression from an established convention - it matches its own template file exactly. Would be a nice improvement, but holding only the new file to a stricter i18n bar than the page it's copied from seems inconsistent. Not blocking on this one.

Requesting changes for #1 (real security gap, matches an actively-enforced pattern elsewhere in the codebase) - the other two are worth a look but not required before merge.

LuiHedlund02:
@ZaynSaul Addressed your blocking point and rebased the branch.

**1. `returside` sanitization (blocking) — fixed.** `lager/lister/styklister.php` now routes the request value through `nav_sanitize_returside()` before it reaches `topLineVarer.php`, using the same call shape as `lager/varekort.php` / `lager/varer.php`:

```php
$returside = nav_sanitize_returside(if_isset($_GET, "", "returside"));
if ($returside === "") {
    $returside = get_relative() . "index/menu.php";
}
```

Rejected and missing values fall back to `index/menu.php`, so the previous default is preserved. Verified against the payload set from `tests/characterization/includes/NavReturnTargetRegressionTest.test.php`: `javascript:alert(1)`, `javascript&#58;alert(1)`, `x onclick=alert(1)`, `https://evil.tld/x` and `returside[]=x` all fall back to the menu; a legitimate `lister/vareliste.php` passes through unchanged.

**2. `momspris` duplication — left as is, deliberately.** There is no existing home for it on master: `includes/stdFunc/` holds PHP behaviour helpers, not SQL fragments, and the same `CASE` is already duplicated across `lager/lister/vareliste.php`, `lager/lister/ordrestatus.php` and `lager/lister/indkøb.php`. Extracting it would mean inventing a new pattern and touching three unrelated files in this PR. Worth a separate ticket that does all four call sites at once.

**3. Danish headers — not changed**, agreed with your reasoning.

Also merged current `origin/master` into the branch (it was 7 months behind). The merge was conflict-free; only the two PR files differ from master. I also reverted the incidental trailing-whitespace change on the neighbouring Serienumre block so the diff stays scoped.

Verification: `php -l` clean on both touched files; `NavReturnTargetRegressionTest` passes (35/35) under PHPUnit 11.5.56 / PHP 8.3.33. Not verified: no browser run of the page against live data.

ZaynSaul (approved):
Verified. styklister.php now runs the GET-supplied `returside` through `nav_sanitize_returside()` before use, falling back to `index/menu.php` when it's rejected/empty (matches the function's return-'' contract exactly). Same protection this page's siblings already have. LGTM.

PR #213, opened 2026-02-20
Lagerstatus opdatering

logicguy1:
- Added a proper button to enter the lagerstatus mode, this is much more visible for new users
  image

- Updated the colors to be more pleaseing to the eye and less in your face
  image

- Updated the way that colors and the states are handled to use a central color definition instead of using repeated and duplicated hex values as state management
  https://github.com/DANOSOFT/saldi/pull/213/changes#diff-bac467ef2105a90f895068c804d8227fc7cf416481b21b62d251e96b5c1164c9R176-R181
- Added a highlight in the tooltip that shows what state the current row is in
  Swapped the flow within the tooltip so it goes
  - NOT IN STOCK
  - STOCK ORDERED
  - IN STOCK
  - READY TO PRODUCE INVOICE
  
  It is a more natural flow than what was there before
  - IN STOCK
  - READY TO PRODUCE INVOICE
  - STOCK ORDERED 
  - NOT IN STOCK
  
  image

 
- Removed title tag from order numbers when in lager status mode (this was blocking the lager status popup)
- Fixed translation key for magenta (had gotten overwritten)
- Added a back button to exit the lagerstatus view, before the only way to exit was to go to offers and go back to orders


LuiHedlund02:
@logicguy1: still current? `debitor/ordreliste.php` has 16 master commits since June, so a real rebase is needed. Note the `title='$title'` removal on the order-number cell is unconditional, not only in lagerstatus mode.
Lead reviewer once rebased: @ZaynSaul.

LuiHedlund02:
@logicguy1 please rebase onto current master before @ZaynSaul’s stock-status presentation review. The diff removes title text outside lagerstatus mode too; preserve ordinary tooltips or document/test why that change is intended. Include before/after checks for stock-status and ordinary order-list modes.

ZaynSaul (changes requested):
Confirmed both of Lui's points:

**Rebase needed** - GitHub reports this PR as unmergeable (`mergeable_state: dirty`), consistent with the 16 master commits on `debitor/ordreliste.php` since June that Lui mentioned.

**Unconditional title-tag removal - confirmed real bug.** Traced the diff: `return "<td align='$column[align]' style='$style' $onclick title='$title'>$display</td>";` becomes `return "<td align='$column[align]' style='$style' $onclick>$display</td>";` - and this return statement is the closure's final return, sitting *outside* the `if ($vis_lagerstatus && ...)` block that wraps the new tooltip-building logic above it. So this removes the `title` attribute from every order-list row in *every* mode, not just lagerstatus mode. The PR description says the intent was only to stop it blocking the lagerstatus popup - as written, ordinary (non-lagerstatus) order list rows lose their tooltip too.

Please rebase onto current master and scope the title removal to `$vis_lagerstatus` mode only (e.g. keep `title='$title'` in the non-lagerstatus branch, or only drop it when the overlib span tooltip is actually being attached). Happy to re-review once that's in.

LuiHedlund02:
@ZaynSaul Both review points are addressed in 866d8f04.

**Rebase.** Merged current `origin/master`; `debitor/ordreliste.php` was the only conflicted file. Master had refactored the lagerstatus computation into `get_order_lagerstatus_cache()` (shared by `rowStyle`, the tooltip render callback and the bulk `preload`), so I re-applied the PR's hunks onto master's version of the file rather than the reverse. Master's behaviour wins everywhere the PR did not intend to change something — in particular master's `<a href>` row link and the bulk preload path are untouched.

Notable consequence: the PR's hand-rolled `$overall_bg` aggregation is gone, replaced by `$cached_ls['color']` from the shared helper. Same priority order (red > yellow > green > magenta), but the status icon and the row background colour now come from one source and cannot disagree. The LS_* constants are also applied to the shared helper and the preload block, so there are no hardcoded hex colours left in the file.

Also carried over from the PR: `findtekst('2403|Alt leveret')` -> `findtekst('1425|Alt leveret')`. Text id 2403 in `importfiler/tekster.csv` is "Opret en ny vare"; 1425 is "Alt leveret, mangler kun at fakturere."

**Title tag.** Scoped. The closure now sets `$lagerstatus_tooltip = true` only inside the branch that actually wraps `$display` in the overlib span, and the final return builds the attribute from it:

```php
$td_title = $lagerstatus_tooltip ? "" : " title='$title'";
return "<td align='{$column['align']}' style='$style'$td_title><a href='$href' ...>$display</a></td>";
```

Non-lagerstatus rows keep their tooltip; `art == 'DK'` (kreditnota) rows keep it too, since they are excluded from the popup.

**Verified:** `php -l` clean on `debitor/ordreliste.php`; `tests/integration/printBatchHandoff.php` passes (2/2).

**Not verified — needs a browser:**
- The lagerstatus popup actually rendering, and the correct legend row being highlighted, for each of the four statuses.
- The four SVG icons resolving at `../ikoner/` from `debitor/ordreliste.php` and looking right at 20x20 next to the order number.
- Restordre rows (`restordre == '1'`) now getting a popup at all — this PR removed that exclusion, and master's preload path was written while the exclusion was still in place.
- Native title tooltip vs overlib popup on the same page, i.e. that ordinary rows still show the "Fortsæt med at redigere ordren" tooltip while lagerstatus rows show only the popup.
- The new Vis lagerstatus / Tilbage buttons rendering correctly in `#left-controls` — master changed the surrounding separator from two spaces to ` | `, and the buttons no longer participate in that separator scheme.
- `tests/characterization/debitor/OrderEmployeeSearchTest.test.php` could not be run: PHPUnit is not installed in this environment (no `vendor/`). Pre-existing, unrelated to this change.

ZaynSaul (approved):
Verified the review fix directly: the native `title` attribute is now only dropped on cells where `$lagerstatus_tooltip` is actually true (set only when the overlib span is built), so the earlier unconditional-removal issue is gone — regular list cells keep their row title again. Restordrer are now included in the tooltip as intended, and the 2403→1425 findtekst correction for 'Alt leveret' is right (matches the legend label going through the same key). Centralized color constants and the icon map read cleanly. LGTM.

ZaynSaul:
Small follow-up for clarity, per the same testing-limits standard I'm applying across my open reviews: my approval verified the two things I'd previously flagged (rebase/mergeability, and the unconditional `title`-attribute removal now correctly scoped to lagerstatus rows) at the code level, and that verification holds.

To be explicit about what it doesn't cover: the author's own handoff comment lists several items as "not verified — needs a browser" (popup rendering per status, icon resolution, restordre rows actually showing a popup, native-tooltip vs. popup coexistence). My approval didn't restate that those remain unverified — it wasn't meant to imply they'd been checked. Flagging here so it's explicit before merge/deploy: those are still author-owned, browser-only checks.

SirRolin (approved):

PR #256, opened 2026-04-10
Return delivery

PatrickMadsen:


kreditor/orderIncludes/openOrderLines.php
```diff
@@ -72,7 +72,11 @@
   print "<td title='".findtekst(1513, $sprog_id)."'><input class='inputbox' type='text' style='background: none repeat scroll 0 0 #e4e4ee' readonly=readonly size=7 name=vare$x onfocus='document.forms[0].fokus.value=this.name;' value='".htmlentities($varenr[$x])."'></td>"; #20180305
   print "<td><input class='inputbox' type='text' size=7 name=lev_varenr$x value=\"".htmlentities($lev_varenr[$x])."\" ";
   print "onfocus='document.forms[0].fokus.value=this.name;' onchange='javascript:docChange = true;'></td>";
-  print "<td><input class='inputbox' type='text' style='text-align:right' size=4 name=anta$x value='$dkantal[$x]' "; print "onchange='javascript:docChange = true;'></td>";
+  if ($fokus == 'anta' . $x) {
+    print "<td><input class='inputbox' type='text' style='text-align:right' size=4 name=anta$x placeholder='$dkantal[$x]' value='' onchange='javascript:docChange = true;'></td>";
```
SirRolin, line 76: Doesn't value='' mean that if you then change some other tabledata and save that antal again becomes 0?

LuiHedlund02:
@PatrickMadsen: this is now a two-line uncomment that re-enables the `levering.php` redirect on negative `leveres`. @SirRolin's question from 2026-05-08 is unanswered and the description is still the template. Please state why the block was commented out originally and what changed, or close the PR.
Lead reviewer if it stays: @SirRolin (order lifecycle, `batch_salg`).

LuiHedlund02:
@PatrickMadsen please answer @SirRolin’s May question on the return-delivery behavior and update the PR description with the intended change and verification. If the work is no longer needed, please close it; it is waiting on the author rather than a new reviewer.

ZaynSaul (changes requested):
Checked the *actual* current diff via the API (not `gh pr diff`, which shows the full historical per-commit patches including content that's since been superseded) - it's now just 2 deletions in debitor/ordre.php: removing the `/* ... */` wrapper around the negative-leveres → `levering.php` redirect block.

Good news: I traced @SirRolin's May question ("Doesn't value='' mean... antal again becomes 0?") to the `kreditor/orderIncludes/openOrderLines.php` change from the original commits - that file isn't part of the current diff anymore, so his specific concern no longer applies to what's actually being proposed now.

But that doesn't resolve the real blocker: current master already has the negative-leveres clamping this PR originally tried to add (`debitor/ordre.php:2097`, dated 20260415, after this PR's commits) and the `$art=='DO'` batch_salg handling in `includes/ordrefunc.php:371` - all merged independently since. The only thing left is un-commenting a redirect that someone deliberately disabled at some point, with no comment explaining why. Re-enabling disabled code without knowing why it was turned off is a real risk regardless of how small the diff looks now - a redirect that yanks the user to `levering.php` mid-save could have been disabled because of exactly that kind of disruption for a specific workflow.

Same ask as Lui's: please state why this was originally commented out and why it's safe to re-enable now (or close if this is no longer needed) - this is genuinely waiting on author context, not further review.

PR #284, opened 2026-05-18
Sd 244

SirRolin:
Fixed Negative Deliveries returning delivered to 0 instead of the intended amount-.

Fixed SSL12 trying to use Locator.php, a live service only.

Stopped $_SERVER['SERVER_NAME'] from being appended $_SERVER['PHP_SELF'] on Locator.php check.


importfiler/tekster.csv
```diff
@@ -3053,7 +3053,7 @@
 3053	Sagsadresse	Project address	Saksadresse
 3054	Med. nr.	Emp. no.	Med. nr.
 3055	Medarbejdernavn	Employee name	Medarbeidernavn
-3056	Transport	Transport	Transport
+3056	Transport	Transport	TransportLagerhistorik	Inventory history	Lagerhistorikk
```
cph-pk: There is double text here.
I think the last three should be there

cph-pk (changes requested):

importfiler/tekster.csv
SirRolin: The "lagerhistorik" and such was added later in the file, this was a merge error.

cph-pk (approved):
Looks nice now

PR #315, opened 2026-06-03
HML-303-direct merge into prod

SirRolin:
Pulled in SD 510 & SD 519 from pr#303 and fixed the errors.

removed pos_ordre.php extra random line that broke brackets.


cph-pk (approved):
The code looks fine.

PR #317, opened 2026-06-05
Fix: false unsaved popup when creating new creditor order

ZaynSaul:
## Summary
After saving a creditor order and pressing "New" to create a new 
one, a popup incorrectly appeared with the message "is not saved". 
Since the current order had already been saved, this warning should 
not appear.

- Fixed the saved state tracking so it correctly resets after 
  a successful save
- Ensured the "New" button does not trigger the unsaved changes 
  popup when the order has already been saved

## Files Changed
- kreditor/ordre.php
- includes/kreditorOrderFuncIncludes/topLine_S.php

## How to Test
1. Log in to test_7 (Rotary)
2. Navigate to Creditors → Orders
3. Create a new creditor order and fill in the required fields
4. Save the order
5. Press "New" to create another creditor order
6. Verify no "is not saved" popup appears
7. Verify the popup still appears correctly when there ARE 
   unsaved changes and user tries to navigate away

## Acceptance Criteria
- No false "is not saved" popup after saving and pressing New ✅
- "is not saved" popup still appears correctly for unsaved changes ✅
- Fix verified on test_7 (Rotary) ✅
- No other order creation or navigation flows affected ✅

## Tested On
- test_7 (Rotary) ✅

- [ ] I have checked that i am not linking to any external resources such as external style- or javascript-files, that could compomise stabilaty

### DIFF
--- a/includes/kreditorOrderFuncIncludes/topLine_S.php
+++ b/includes/kreditorOrderFuncIncludes/topLine_S.php
@@ -64,7 +64,7 @@ print "<!DOCTYPE html PUBLIC \"-//W3C//DTD HTML 4.01 Transitional//EN\"><html><h
 	if (($kort!="../lager/varekort.php" && $returside != "ordre.php")&&($id)) {
 		
 		  print "<td width=\"5%\">$color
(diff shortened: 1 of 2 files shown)


cph-pk (approved):
This code looks fine.

PR #323, opened 2026-06-11
SD-523 Debtor/Order - Drop-down field to Order

Loe4g:


cph-pk (approved):
This code looks fine.

PR #346, opened 2026-06-27
[codex] Integrate Chaty chatbot widget

LuiHedlund02:
﻿## What changed

- Added a shared Chaty widget include with the SALDI production widget configuration.
- Included the widget immediately before `</body>` in the shared topmenu, report, and old-design footer variants.

## Why

SALDI can load the chatbot through the plain external JavaScript snippet without adding npm, React, Next.js, or any application build step.

## Validation

- `git diff --check`
- Verified include target exists for each updated footer file.
- Browser-checked a local rendered page containing the widget markup and confirmed the script URL, async flag, and expected `data-*` attributes.

Note: `php -l` could not be run because PHP is not installed on this machine's PATH.

### DIFF
--- a/includes/chaty_widget.php
+++ b/includes/chaty_widget.php
@@ -1,0 +1,8 @@
+<script
+    async
+    src="https://wuweiworkai.com/chaty-v2/widget.js"
+    data-widget-id="saldi-prod"
+    data-brand="SALDI.dk"
+    data-lang="da"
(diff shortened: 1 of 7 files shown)


LuiHedlund02:
Closing as superseded by merged #546. Keep the integrated single-loader implementation rather than another widget insertion.

PR #382, opened 2026-07-08
Fix kreditor order split producing blank page

momocoder14:
- Guard all four POST arrays with isset() to handle disabled form inputs
- Fetch antal and leveret from DB in move loop to fix index mismatch (splitOrder uses 1-indexed fields, ordre.php loop sets 0-indexed), which caused last order line to always trigger partial split with zero antal
- Wrap SELECT setval() calls in PostgreSQL-only condition to prevent fatal db_modify failure on MySQL

### DIFF
--- a/kreditor/orderIncludes/moveOrderLines.php
+++ b/kreditor/orderIncludes/moveOrderLines.php
@@ -4,7 +4,7 @@
 //               \__ \/ _ \| |_| |) | | _ | |) |  <
 //               |___/_/ \_|___|___/|_||_||___/|_\_\
 //
-// --- kreditor/orderIncludes/moveOrderLines.php --- lap 5.0.0 --- 2026-06-02 ---
+// --- kreditor/orderIncludes/moveOrderLines.php --- patch 5.0.0 --- 2026-07-08 ---
 // LICENSE
 //
 // This program is free software. You can redistribute it and / or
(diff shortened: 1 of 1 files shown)


kreditor/orderIncludes/moveOrderLines.php
```diff
@@ -109,8 +120,10 @@
 			db_modify($qtxt, __FILE__ . " linje " . __LINE__);
 			$qtxt = "INSERT INTO ordrelinjer SELECT * FROM temp_table WHERE id='$newLineId'";
 			db_modify($qtxt, __FILE__ . " linje " . __LINE__);
-			$qtxt = "SELECT setval('ordrelinjer_id_seq', $newLineId)";
-			db_modify($qtxt, __FILE__ . " linje " . __LINE__);
+			if ($db_type != 'mysql' && $db_type != 'mysqli') {
```
SirRolin, line 123: Note: mysql and mysqli, the mysql and mysqli equivalent is not needed as it's implimented by mysql. This is implimented correctly.

SirRolin (approved):

SirRolin (approved):

kreditor/orderIncludes/moveOrderLines.php
momocoder14: Noted.

PR #403, opened 2026-07-17
Fix data-integrity bugs from ssl3 live: order-line splits, batch updates, posting rollback, invoice-number SAVEPOINT

LuiHedlund02:
## Background

A developer fixed bugs directly on the ssl3.saldi.dk live server without going through GitHub. A full audit compared the live tree against upstream history (every local change was hash-checked against all historical blobs, then adversarially verified against current master). This PR ports the verified data-integrity fixes; master was confirmed to not contain any of them.

## Fixes

- **kreditor/orderIncludes/moveOrderLines.php**
  - The `UPDATE batch_kob SET antal=..., rest=...` statement was built but **never executed** — added the missing `db_modify()` call.
  - Split-off order lines now get `leveret` set explicitly ('0' for a non-delivery split, mirrored for delivered-qty splits) and the original line's `leveret` is updated alongside `antal` instead of going stale.
  - `$bkId/$bkQt/$bkRm/$bkDt` and `$deliveryDate` are reset per order line, so batch state no longer leaks across lines.
- **debitor/bogfor.php** — roll back the open transaction before the error alert/redirect when `bogfor()` returns an error.
- **includes/std_func.php** — `get_next_invoice_number()` uses SAVEPOINT/ROLLBACK TO/RELEASE when called inside an already-active transaction (detected via `pg_transaction_status`), so it no longer prematurely commits the caller's outer transaction during invoice posting. MySQL path unchanged (always own transaction).

## Testing

`php -l` clean on all three files. Recommend exercising: a creditor order-line split with batches, a failing posting (verify rollback), and invoice posting under concurrency on PostgreSQL.

https://claude.ai/code/session_01R1aEDQs2HH2LoHCQ4pQpxg

## Summary by CodeRabbit

* **Documentation**
  * Added an internal change-log note documenting split-line delivery handling, per-line batch resets, and persistence of batch quantity updates.


SirRolin (commented):
Closed due to having no changes after pulling merge in other than a single history line.

PR #411, opened 2026-07-20
Invoice match checkbox

sawaneh1:

## Summary by CodeRabbit

* **New Features**
  * Added configurable initial checkbox selection for popup result rows.
* **Bug Fixes**
  * Improved popup preselection so rows are selected only when all required match conditions are satisfied.
  * Fixed document attachment storage by using the computed document location and automatically creating required folders.
* **Security & Robustness**
  * Hardened attachment path handling via input sanitization and safer directory resolution.


includes/docsIncludes/insertDoc.php
```diff
@@ -338,21 +339,14 @@
     
 
 
-	$path = "../bilag/$db/finance/$kladde_id/$sourceId/";
+	$fileName = basename($fileName); // strip any path/traversal from the filename
+	$path = "$docFolder/finance/$kladde_id/$sourceId/";
```
SirRolin: Due to being imported from a header object ($_REQUEST in this case) we need to sanitise it for security, before using it for any file based functions.
If you could add a function that fails if the realpath() of the target ($docFolder) is not within the realpath of the allowed folders, and use that in this file, it would be ideal.
code example:
`if (($resolvedTarget, $resolvedBase) !== 0) // fail`

SirRolin (dismissed):

SirRolin (dismissed):

SirRolin (dismissed):

SirRolin (approved):

PR #418, opened 2026-07-24
SD-601: Characterization tests for kassekladde posting + order-to-invoice

LuiHedlund02:
Jira: [SD-601](https://saldi-team.atlassian.net/browse/SD-601) — Characterization tests for posting + order-to-invoice (P1)

## What

First automated coverage of the accounting core. **No application code is touched** — only `tests/characterization/` grows.

**`finans/KassekladdePostingCharacterizationTest`** (4 tests) pins `finans/bogfor.php`:
- Simulated posting of a balanced voucher → balanced `simulering` rows, kladde marked `S`, `transaktioner` untouched
- VAT voucher → gross 1250 on an S-coded account splits into net 1000 + 250 VAT posted to the `SM`-group VAT account
- Real posting → balanced `transaktioner` rows, kladde marked `V`, `kontoplan.saldo` moves on both accounts
- Unbalanced voucher → nothing reaches the ledger

**`includes/ordrefunc/OrderInvoiceCharacterizationTest`** (3 tests) pins `includes/ordrefunc.php bogfor($id,'on')` via the production sequence (`remoteBooking/api.php:296-337`):
- Invoicing assigns a `fakturanr`, ends at status 4, keeps sum/VAT, consumes the pending delivery quantity
- Invoice numbers are sequential across orders
- Re-invoicing is refused (current wording pinned) with `fakturanr` unchanged

## How (the interesting part)

`finans/bogfor.php` is a page script (session, relative includes, `exit()` branches) — it cannot be included inside PHPUnit. The harness:

- **Clones a throwaway tenant per run** (`CREATE DATABASE saldi_chartest TEMPLATE saldi_2`), registers it in master `regnskab`, and seeds an `online` session row (`rettigheder` is a positional digit string — `includes/online.php:331`).
- **Drives the page in a child PHP process** through its own `$_POST['simuler']/['bogfor']` dispatch — the exact production path including `transaktion('begin')/('commit')`.
- **Bootstraps `ordrefunc.php` the way `remoteBooking/api.php:29-60` does in production** (master connect → `$db` → include → tenant reconnect → derive `$regnaar`).

Tests assert on database state, never on page HTML. See [tests/characterization/README.md](../blob/sd-601-posting-characterization/tests/characterization/README.md).

## Verified

| Where | Result |
| --- | --- |
| Inside docker web container | **33 tests, 82 assertions, green** (26 usdecimal + 7 new) |
| Host without the stack | 26 pass, 7 **skip cleanly with a reason** — `composer test` stays green on a bare checkout |
| Deliberate regression (simulate dispatch flipped to post for real, then reverted) | **2 tests fail** — regression detection proven |

## Running

```bash
docker compose up -d   # installed stack with a tenant (see README)
docker compose exec -T -w /var/www/html/saldi web php vendor/bin/phpunit --testsuite characterization
```

Discovered by `phpunit.xml`, so this rides along with `composer test` (SD-593) and is ready for the CI gate (SD-591) once that job runs against the docker stack.

## Notes for review (Saul — domain check appreciated)

- Status after invoice+delivery is **4**, not 3 — the guard at `includes/ordrefunc.php:1231` only requires `>2`. Pinned as observed; if 4 is itself a bug, the test gives us the tripwire to change it deliberately.
- `levering()` consumes `leveres` back to 0 and does **not** stamp `leveret` on this path — also pinned as observed.
- Fixture accounts/VAT groups are looked up dynamically from the tenant's kontoplan/grupper, so the suite works on any standard-seeded tenant.

## Summary by CodeRabbit

* **Tests**
  * Added characterization coverage for simulated and real ledger posting, including VAT handling and rejection of unbalanced entries.
  * Added coverage for order-to-invoice conversion, sequential invoice numbering, and prevention of duplicate invoicing.
  * Added a documented test environment that skips database-backed tests cleanly when PostgreSQL is unavailable.
  * Added verification guidance and expected failure behavior for intentionally broken posting flows.


LuiHedlund02:
@SirRolin added disposable-tenant teardown (DB plus scoped chartest registry/session rows), TEST_DB != TEMPLATE/MASTER guards, environment-only DB password, child exit assertions, and full error visibility. Template sessions are no longer terminated. Two harness-safety tests pass (5 assertions); seven DB-backed posting tests skip without SALDI_CHAR_PGPASS and remain unverified here. Existing literal test credentials must be rotated if reused. Please review the harness and answer the three behavior questions before treating it as a contract: (1) should posted orders end at status 4? (2) is consuming pending delivery without stamping `leveret` intentional? (3) is the sales VAT account/rate from the SM group the intended pinned behavior?

ZaynSaul (approved):
Confirmed CodeRabbit's db-collision safety guard is properly implemented: `assertSafeDatabases()` is called as the very first line of `bootstrapTenant()`, before the `pg_terminate_backend`/`DROP DATABASE` calls, and also gates `teardownTenant()`. Good, cheap protection for a destructive operation.

On the two domain questions:

**Status 4 instead of 3 after invoice+delivery** - checked the other status-setting sites in `ordrefunc.php`: line 1591 sets status='3' on a plain invoicing path, while lines 2237/3164/3262 set status=4 specifically alongside `report_number` (the POS/cash-register report-closing path). Since this test drives the flow through `remoteBooking/api.php`'s production sequence, landing on status=4 looks like it's hitting the report_number-linked path rather than a bug - these look like two legitimately different terminal states for two different invoicing entry points, not an off-by-one. Pinning it as observed rather than asserting status=3 is the right call either way.

**`levering()` not stamping `leveret`** - checked every consumer of `ordrelinjer.leveret` I could find (`lager/rapport.php`, `mysale/api.php`, `includes/ordrefunc.php`'s own outstanding-quantity calculations) - they all filter to `status<3`. Since this flow's endpoint is status 4, it's excluded from every `leveret`-dependent query I found, so the gap looks low-impact for this specific path. Wouldn't swear there's no other status-agnostic consumer, but nothing I found breaks. Worth a follow-up ticket to confirm, not a blocker on the test itself.

Test-only PR, no application code touched, pins real (not guessed) behavior with the harness properly isolated (throwaway cloned tenant, guarded against a master/template collision). Approving.

PR #433, opened 2026-07-29
SD-619 [P2] Fix stock pages returning HTTP 500 errors

Loe4g:
https://github.com/DANOSOFT/saldi/blob/e6ddbb2379c6e2bf0713235b2ac0c7d90f2518fb/PR.md

From the visible codebase, the `Produktion` directory should be deleted entirely. It is outdated, not initiated, and is not used or referenced by any part of the application.

Removing the directory will make comment review item  1 and the referenced PR **400** stale.

This commit therefore removes the entire `Produktion` directory and all of its contents. The functionality appears to be obsolete, was never fully implemented, and was never integrated into the application.

Comment review items 2, 3, and 5 have been updated accordingly.

For comment review item 4, the key reasons for removing `vareimport.php` are not that it is directly proportional to `systemdata/importer_variantvarer.php`, but that both follow a similar overall workflow structure, suggesting that `vareimport.php` may be a trial MVP or an obsolete counterpart.

There are also similarities such as the identical delimiter auto-detection block. However, the primary justification for removing `vareimport.php` is the file's own status within the codebase:

1. It is not used or referenced anywhere. If it is intended to support a separate feature in the future, it should first be properly developed and exposed through the appropriate UI.
2. The code is broken and outdated, including on live accounts.
3. There is no UI access to the functionality.
4. The file primarily references itself and does not appear to participate in an active application workflow.

Based on these findings, removing `lager/vareimport.php` is appropriate unless there is a concrete plan to develop, integrate, and expose the functionality as an active feature.


LuiHedlund02:
Good news first: this targets `next` (no auto-deploy) and CI is green. But
five things block it:

1. **Decimal behavior change:** `produktion/ordre.php` switches to
   `stdFunc/usDecimal.php`, which is not equivalent — `1,234` now parses as
   1.234 (was 1.23) and `0,004` as 0.004 (was 0.00). Quantities, prices, and
   discounts can persist differently. Either restore two-decimal
   compatibility or get the new precision explicitly approved and tested.
2. **The 500s aren't fully fixed:** `minmaxstock.php` initializes only item
   arrays — empty enabled stock groups/locations/stock rows still leave
   `vGr`/`stk`/`stock` undefined and can TypeError back to a 500.
3. **XSS:** the new `debitor/udlign_openpost.php` missing-parameter redirect
   interpolates raw request values into a meta-refresh attribute — validate
   and use a safe server-side redirect.
4. **Functional loss:** deleting `lager/vareimport.php` isn't covered by
   `systemdata/importer_variantvarer.php` (variant importer ≠ general
   supplier/product-group/discount import). Preserve the workflow or prove
   parity.
5. **License headers:** the GPL/non-compete header rewrites must come out —
   that's a material licensing change nobody has approved.

Also: coordinate with PR 400, which adds a parser + tests for the
`includes/usdecimal.php` you're deleting — PR 400 lands on master first, and
this PR should rebase over it and update `.clanker/verification.yaml`
(it still registers the deleted module as characterized).

Loe4g:
This has been fixed has been fixed since August 12th :)

PR #439, opened 2026-07-31
Sd 513 upgrade

sawaneh1:

## Summary by CodeRabbit

* **New Features**
  * New customers can receive the lowest available account number automatically.
  * Added CVR and phone lookup support with localized status and error feedback.
  * Added optional automatic CVR lookup on customer forms, including improved handling of delayed or failed searches.
  * Added a mobile phone field for employees.
* **Bug Fixes**
  * Improved duplicate account-number handling during customer creation and editing.
  * Prevented invalid or incomplete lookup requests and improved service error handling.
  * Improved account-manager filtering based on permissions.
  * Added safer handling when customer account information is missing.


LuiHedlund02:
The proxy/authentication hardening looks good. Two fixes before merge:

1. The AJAX success callback doesn't check it's still the active lookup —
   a slow older CVR/phone response can overwrite the fields after a newer
   lookup returns. Track a request sequence/abort the stale one.
2. DB-backed translations are inlined with plain `json_encode()` — a value
   containing `</script>` breaks out of the script block (stored XSS). Use
   `JSON_HEX_TAG | JSON_HEX_AMP | JSON_HEX_APOS | JSON_HEX_QUOT`.

Please also test the reversed-response-order case explicitly. (Non-blocking:
the account-number locking is PG-only with no DB uniqueness constraint —
worth a follow-up ticket.)

SirRolin (approved):

PR #444, opened 2026-08-03
Fix ffdage1 grace period being ignored when generating dunnings

sawaneh1:

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected first reminder deadline calculations to use the current date and configured grace period.
  - Ensures overdue invoices are selected consistently for reminders.

- **Tests**
  - Added comprehensive coverage for overdue invoices, grace periods, exclusions, fees, debtor grouping, and reminder escalation.
  - Added shared database-backed test infrastructure and fixtures for reliable characterization testing.


LuiHedlund02:
The four-line production fix is exactly right — it removes the unset-`$dd`
root cause of the grace-period bug, and I want it in fast. The blockers are
all in the bundled test infrastructure (~1,300 lines), so my ask is to
split:

1. **Keep in this PR:** the `ny_rykker.php` fix — but switch the cutoff
   from 86400-second arithmetic to calendar-day arithmetic; the current form
   shifts dunning eligibility by a day around the Copenhagen DST
   transitions. Add a test for `ffdage1+1` and a DST boundary.
2. **Move to PR 418's harness:** the CharacterizationEnv/fixtures work.
   As written it will DROP DATABASE any env-supplied name with no guard
   against master/template/real tenants, and it leaves a predictable
   all-rights `online` session in the master DB after teardown. The same
   harness appears in PR 418 and PR 456 — we're hardening it once, in 418.

With the split + DST fix this merges quickly.

sawaneh1:
> The four-line production fix is exactly right — it removes the unset-`$dd` root cause of the grace-period bug, and I want it in fast. The blockers are all in the bundled test infrastructure (~1,300 lines), so my ask is to split:
> 
>     1. **Keep in this PR:** the `ny_rykker.php` fix — but switch the cutoff
>        from 86400-second arithmetic to calendar-day arithmetic; the current form
>        shifts dunning eligibility by a day around the Copenhagen DST
>        transitions. Add a test for `ffdage1+1` and a DST boundary.
> 
>     2. **Move to PR 418's harness:** the CharacterizationEnv/fixtures work.
>        As written it will DROP DATABASE any env-supplied name with no guard
>        against master/template/real tenants, and it leaves a predictable
>        all-rights `online` session in the master DB after teardown. The same
>        harness appears in PR 418 and PR 456 — we're hardening it once, in 418.
> 
> 
> With the split + DST fix this merges quickly.

The Split is done.

Split done: harness dropped (lives in #418), cutoff switched to calendar-day arithmetic via dunning_cutoff_date(), tests for ffdage1+1 and both DST switches added. I applied the same change to the ffdage2/3 cutoffs since they had the identical formula

LuiHedlund02 (approved):

PR #445, opened 2026-08-03
Sz/fix/tinymce upgrade dedupe

ZaynSaul:
## Summary
`tiny_mce/tiny_mce.js` was TinyMCE 3.5.8 from 2012 — roughly 14 years past end-of-life, with
known XSS fixes only shipped in later majors. It was actively loaded by 3 pages
(`sager/tilbud.php`, `sager/notat.php`, `sager/template_form.php`) that let users type
free-form rich text saved straight into the accounting database — a realistic stored-XSS
vector. There was also a byte-identical, completely unreferenced second copy of the whole
library at `sager/tiny_mce/`, left over from a historical bulk server migration.

This PR upgrades to TinyMCE 8.8.2, updates all call sites to the new API, and removes the
dead duplicate copy.

1. Deleted `sager/tiny_mce/` — confirmed dead via repo-wide grep (zero references) and
   matching git history (both copies were introduced in the same old migration commit).
2. Vendored TinyMCE 8.8.2 into `tiny_mce/` (replacing the v3 files in place) — the only
   major version still receiving open-source security support; v5/6/7 are already EOL.
   Added `license_key: 'gpl'` to each `tinymce.init()` call, which self-hosted v6+ requires
   to run without a nag.
3. Updated all 3 call sites to the new script filename and `tinymce.init()` API, rebuilding
   the plugin/toolbar lists to v8-native names and dropping what's genuinely gone (merged
   into core, or premium-only with no free equivalent), with `browser_spellcheck: true`
   replacing the old server-side spellchecker plugin.
4. Ported the `-mytemplate` custom plugin — a toolbar dropdown that inserts saved
   case-boilerplate text from the `sagstekster` table. This initially looked like a dead
   reference during exploration, but turned out to be a working feature; rewrote it using
   v8's `editor.ui.registry.addMenuButton` API instead of dropping it.
5. Removed TinyMCE's "Get all features" premium upsell banner (`promotion: false`) —
   inappropriate for an internal accounting tool, noticed during live testing.
6. Added a file-history line to `notat.php` per repo convention (the only one of the 3 files
   with an existing header block).

## Files Changed
- tiny_mce/ (vendored library, v3.5.8 → v8.8.2)
- sager/tiny_mce/ (deleted — dead duplicate copy)
- sager/tilbud.php
- sager/notat.php
- sager/template_form.php

## Out of Scope
- Migrating away from a rich-text editor entirely.

## How to Test
1. Open each of the 3 editor pages — `sager/tilbud.php`, `sager/notat.php`,
   `sager/template_form.php` — and verify the TinyMCE 8.8.2 editor loads correctly with no
   console errors and no "Get all features" upsell banner.
2. Verify the toolbar/plugin set is functional (formatting, tables, links, etc. as
   applicable per page).
3. Verify browser-based spellcheck works in place of the old server-side spellchecker
   plugin.
4. Open the `-mytemplate` dropdown and confirm it still inserts saved case-boilerplate text
   from `sagstekster` correctly.
5. Do a full create → save → retrieve round trip: type rich text, save, reload the page/
   record, and confirm the content persists correctly through the real database.
6. Confirm `sager/tiny_mce/` no longer exists in the repo and nothing references it.
7. Run `php -l` on all 3 updated PHP files to confirm no syntax errors.

## Acceptance Criteria
|---|---|---|
| 1 | Editor pages (notat, template_form, tilbud) | Function correctly on TinyMCE 8.8.2 (verified in Prodtest) |
| 2 | Repo-wide check for editor library copies | Only one editor copy remains in the repo |

## Verification
- PHP syntax-checked all 3 updated files (`php -l`) ✅
- Tested the vendored library in a headless-browser harness ✅
- Confirmed live against the running app (saldidb2): editor loads, `-mytemplate` dropdown
  works, full create → save → retrieve → insert round-trip through the real database
  succeeded ✅


LuiHedlund02 (commented):
# Review — TinyMCE 3.5.8 → 8.8.2 (SD-604)

**Verdict: LGTM overall, one fix requested before merge.** Multi-agent review (Codex GPT-5.6 + Claude): API-correctness, security, regression, and integration passes, with every finding adversarially verified. Integration checks were run empirically — the vendored build was driven in headless Chromium against faithful reproductions of all three pages.

## Verified clean ✅

- **Editors initialize on all 3 pages.** `#dagbog`, `#temptext`, `#tilbud` all exist in the markup; live runs show zero console errors, working toolbars, and clean asset loads (no 4xx) in both `da` and `en`.
- **v8 API usage is correct.** All plugin/toolbar names valid, every referenced plugin exists in the vendored tree, `langs/da.js` is genuine v8 format, and the `mytemplate` port (`addMenuButton`) works — including on empty/null template lists.
- **No stale references, no v3 leftovers.** Only the 3 intended files reference the library; `sager/tiny_mce/` fully deleted. The dropped `content_css`/list-file URLs pointed at files that never existed (they were 404ing), so removing them is a net fix.
- **Content round-trips.** v8 patches `form.submit()` and syncs editor → textarea before serialization (verified by actual submission).
- **Security: nothing introduced or worsened** by this PR (see notes at the bottom).

## Fix before merge 🔴 (medium)

**The toolbar "Gem" (save-plugin) button silently discards edits** — `sager/template_form.php:171` (same on `sager/tilbud.php:90`). TinyMCE's save plugin calls native `form.submit()`, which does not include a submit button's name/value — but persistence is gated on `if(isset($_POST['template']))`, which only the real `<input type="submit" name="template">` supplies. Verified live: clicking the editor's Gem button POSTs `tekst=...` with **no `template=` key**; the handler falls through and the edit is silently lost. This existed in v3 too, but since the toolbar was rebuilt from scratch here, this is the moment to either drop `save` from plugins/toolbar or wire `save_onsavecallback`.

## Suggestions (low, deliberate-choice)

1. **Editors became vertically resizable** — old configs set `theme_advanced_resizing: false`; v8 defaults `resize: true`. Add `resize: false` if the fixed layouts were intentional.
2. **notat.php sets no `language`** — Danish users get an English toolbar on the dagbog editor while the other two pages show Danish. One-liner, and this PR already ships `da.js`.
3. **No `content_css`** — editor content now renders with TinyMCE's default stylesheet; WYSIWYG fidelity vs. the app is unspecified. Fine to accept, just deliberately.
4. **Vendored weight**: ~10 MB tracked, of which ~5.2 MB is unminified `.js` duplicates + 44 `.ts` files that are never requested at runtime (request logs show only `.min.*` assets load). Deleting non-min sources would halve the footprint. Also: notat.php's editor payload grows ~412 KB → ~1.34 MB uncompressed, and the bundle loads even on read-only note views (`$status >= 1`, no `#dagbog` rendered) — benign but wasteful; the script block could be gated on the same status branch.
5. **Minor feature loss** (probably acceptable, noting for the record): editor print button and the old cite/abbr/attribs/layer controls have no v8 open-source equivalent.

## Pre-existing issues surfaced (out of scope here, worth tickets)

- **SQLi in `sager/template_form.php`**: raw `$_GET['edit']` reaches an unquoted `WHERE id =` UPDATE on `sagstekster` (~line 57 via lines 97-99) before the `is_numeric` check at ~122-125.
- **Stored XSS for viewers is NOT fixed by this upgrade**: finalized note views (`notat.php` ~line 936) still render saved HTML unsanitized; TinyMCE 8's sanitization protects the editor only, not direct POSTs or server-rendered output. Server-side filtering is still needed.
- **`tilbud.php` has no POST handler at all** — its editor content was never saveable, before or after this PR. Fix or retire?

ZaynSaul:
@LuiHedlund02 the "fix before merge" item is addressed in `08cabe19`: dropped `save` from the `plugins`/`toolbar` config in both `template_form.php` and `tilbud.php`, since the toolbar's own Gem button posted without the `template` key and silently discarded edits. The existing "Gem/ret skabelon" submit button remains the only save path.

Left the deliberate-choice suggestions and the pre-existing out-of-scope issues (SQLi in template_form.php, stored XSS in notat.php, tilbud.php's missing POST handler) as-is per your note that those are worth separate tickets — happy to open them if useful.

Would appreciate a re-review when you get a chance.

SirRolin (approved):

PR #447, opened 2026-08-04
Feature/udfoert af

momocoder14:
Udfort af dropdown
Order locking - preventing two users at the same time
Print button fix - no options page
Navigation fix - close returns to card and not overview

## Summary by CodeRabbit

- **New Features**
  - Added order editing protection with conflict alerts, automatic lock renewal, and release when leaving an order.
  - Added a sortable, searchable “Udført af” column to order lists.
  - Added delivery-address selection, saving, clearing, and autofill improvements.

- **Bug Fixes**
  - Corrected order, customer, printing, and back-navigation links.
  - Improved order selection and autocomplete behavior.
  - Corrected the Danish “Udført af” label.

- **Reliability**
  - Improved database updates for older installations and missing tables.
  - Added safeguards for schema variations and migration permission issues.


LuiHedlund02:
go though the review from coderabbit @momocoder14

momocoder14:
> go though the review from coderabbit @momocoder14

On it, I am doing a refactor as well due to the reserved hvem field. The new pull request will feature these changes.

LuiHedlund02:
Holding this one — it's CONFLICTING with master (`debitor/ordre.php`
changed), and the review turned up design issues that a plain rebase won't
fix:

1. **Lock ordering:** the creditor-order lock is acquired inside
   `ordreside()`, *after* the POST dispatcher has already written. A stale
   save can modify the order and only then show a lock conflict. The
   ownership check must run before every mutating path — and
   `clear_delivery` is currently excluded from the early check even though
   it immediately clears persisted delivery fields.
2. **Migration breaks existing consumers:** the migration copies
   `ordrer.hvem` → `performed_by` and blanks `hvem`, but the changed-order
   list still does `SELECT DISTINCT hvem`, and document printing still reads
   `$row['hvem']` for the `ordre_hvem` placeholder in
   `includes/formfunk.php`. After migration, performers vanish from the list
   and print as empty. Every consumer needs to move to `performed_by` in the
   same PR.
3. **Delivery-address persistence isn't transactional:** all primaries are
   demoted first, then replacements written with suppressed `@pg_query` — a
   failed write can leave zero (or multiple) primary addresses while the
   save looks successful.
4. Also: schema migrations are racy read-then-act sequences in the request
   path (including the `record_locks` creation), the fax→mobile rename skips
   installs that have both columns, `includes/grid.php` still emits the
   unescaped search term (the CodeRabbit reflected-XSS finding), and
   `performed_by` accepts non-scalar POST input.

**Ask:** rebase onto current master first (resolve `debitor/ordre.php`
deliberately), then fix 1-3 as a minimum before re-requesting review.
Test plan: new + migrated orders, list filter/sort, configured `ordre_hvem`
print fields, copy/credit/split, delivery-address failure handling, and two
concurrent sessions on both debtor and creditor orders.

SirRolin (changes requested):
After testing the "hvem" column is still broken, in the sense that it's no longer updated when you enter the ordre.
The "hvem" Column is badly named, but it's perpose is to lock the ordre so if you have multiple people on the system, only 1 person can be in a ordre at a time.

If we were to rename the column, it should be renamed to "locked_by", but that requires checking all ordre files for "hvem" uses.

LuiHedlund02:
@momocoder14 decision: #575 will carry the remaining optional performer/translation fix using the renderer already on master (ID 5231), preserving the existing concurrent-edit warning. Please separate this PR’s competing `record_lock.php` / release / heartbeat mechanism from the performer UI and reconcile its design with #589 before asking for review. If any unique performer behavior remains, identify it so we can preserve it without another renderer or lock mechanism.

momocoder14:
Closing this PR as its remaining work has now been split/superseded by the newer changes.

The remaining Udført af / performer behavior is being handled by #575 using the renderer already on master, while the concurrent-edit locking work should be reconciled with #589 rather than maintaining a competing lock/release/heartbeat implementation here.

Three fixes from this PR are not carried forward and are still open on master — the unescaped search term in includes/grid.php (reflected XSS, CodeRabbit flagged it here on 2026-08-04), the double ? in debitor/udskriftsvalg.php's redirect, and the non-transactional delivery-address save in debitor/debitorkort.php (Lui's review item 3). I'm taking those out as separate PRs off master so they don't close with this one; links to follow.

PR #449, opened 2026-08-06
Feature/konto udtog

momocoder14:
Kontoudtog added to forms and visual editor

## Summary by CodeRabbit

* **New Features**
  * Added the `konto_udtog` field for account statements, with Danish and English labels.
  * Account statements now display outstanding balances from unsettled transactions.
  * Added a sample preview value for the new account-statement field.

* **Localization**
  * Added localized field options for account-statement forms.

### DIFF
--- a/includes/formfunk.php
+++ b/includes/formfunk.php
@@ -644,7 +644,18 @@ if (!function_exists('find_form_tekst')) {
 						$db_variabel = ($variabel == 'cvr') ? 'cvrnr' : (($variabel == 'fax') ? 'mobile' : $variabel);
 						$q2 = db_select("select $db_variabel as $variabel from adresser where art='S'", __FILE__ . " linje " . __LINE__);
 					} elseif ($tabel == "adresser" || $tabel == "konto") {
-						if ($variabel == 'valuta') {
+						if ($variabel == 'udtog') {
+							// 20260805 MJ konto_udtog: udestående saldo fra openpost
+							if ($formular == 11) {
+								$kid = (int)$id;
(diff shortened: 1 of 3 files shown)


LuiHedlund02:
The field wiring (editor + form renderer) looks right. One blocker, which
is also CodeRabbit's open comment: the balance in `includes/formfunk.php`
uses `sum(amount)` across openpost rows, ignoring each row's
`valuta`/`valutakurs` — for accounts with open posts in multiple currencies
the printed outstanding balance is wrong. Please accumulate per row using
the same conversion approach as skyldig/forfalden, then format the
converted total. A quick regression with one DKK + one EUR open post on the
same account should prove it.

LuiHedlund02:
Deferred account-statement/currency work; needs domain review and a resolved merge order for formfunk.php. Marked `udskudt` so the new stale workflow cannot close deliberately deferred work. No reviewer request is being added while parked.

ZaynSaul (approved):
Traced the new konto_udtog field through formfunk.php, formeditor.php and formularkort.php - it's a genuinely new field (sums openpost for udlignet='0' rows, DKK-converts the same way skyldig/forfalden already do), both id sources going into the SQL are forced numeric before interpolation, and the output goes through dkdecimal/afrund before it ever reaches the print template, so no escaping gap. Translation key 1803 it references already exists.

Only thing missing is a history-block line in formularkort.php even though it's actively modifying drop_down() there - small thing, doesn't block this from my side, just flagging for whenever you're back in that file.

PR #450, opened 2026-08-06
Fix/openpost report global padding

sawaneh1:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved open-post report display and layout consistency.
  * Fixed report behavior when open posts are hidden, preventing malformed page markup.
  * Preserved view, pagination, and PBS settings when navigating or submitting reports.
  * Improved table alignment, column sizing, scrolling, and pagination controls.
* **Performance**
  * Simplified report loading by removing the intermediate loading frame.


LuiHedlund02:
Deferred open-post report change; resume after #620 and rebase the overlapping report code. Marked `udskudt` so the new stale workflow cannot close deliberately deferred work. No reviewer request is being added while parked.

ZaynSaul (changes requested):
The bug is real and still there in master today - colspan=10 is still hardcoded in a few spots in showOpenPosts.php even though the correct computed value already exists right next to it, so the fix is legitimate, and it's a genuinely shared function between debitor and kreditor, not a debitor-only patch leaking into kreditor's rendering.

Two things stopping approval: this PR is conflicting with master right now - showOpenPosts.php has grown a lot since this branch point (CSV export, aging-bucket filtering, balance rule changes all landed since), so it needs a rebase and the fix re-verified against the current file before it can merge. Separately, while checking the escaping fix in rapport.php I found the identical pattern one line up - line 288 has the exact same unescaped $husk/$dato_fra/etc. interpolation that got fixed at line 305, just missed. Since you're already touching this exact db_modify pattern in this file, worth fixing both while you're in there rather than leaving a live SQL injection surface right next to the one that got closed.

PR #451, opened 2026-08-07
Fix SST-715: urlencode returside i udskriv-redirect fra formfunk.php

momocoder14:
returside blev indsat raa i redirect-URL til udskriv.php. Naar brugeren soeger paa kundenavn i ordrelisten indeholder returside-URL search- parametre som f.eks. &valg=faktura, der laekker ud som selvstaendige GET-parametre i udskriv.php. Dette medforer at $valg saettes til 'faktura' i stedet for 'pdf', PDF-fremviseren springes over, og brugeren ender paa en forkert 'Udskriftsvalg'-side.

Urlencode() paa $returside sikrer at & og ? i vaerdien er korrekt kodet og ikke fortolkes som URL-separatorer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form redirects by safely encoding return-page information, helping ensure users are sent back to the correct location after printing.
  * Restricted return-page links to relative paths and added safeguards for navigation links, preventing unsafe redirects and injected content.

### DIFF
--- a/includes/formfunk.php
+++ b/includes/formfunk.php
@@ -2260,7 +2260,8 @@ if (!function_exists('formularprint')) {
 		} elseif ($nomailantal > 0) {
 			print "<big><b>Vent - Udskrift genereres</b></big><br>";
 			$mappe = str_replace('../temp/', '', $mappe);
-			print "<meta http-equiv=\"refresh\" content=\"0;URL=../includes/udskriv.php?locat=$locat&ps_fil=$mappe/$printfilnavn&amp;id=$id&amp;udskriv_til=$udskriv_til&amp;art=$art&amp;bgr=" . urlencode($background_pdf_path) . "&returside=$returside\">";
+			// 20260807 MJ urlencode returside saa search-parametre ikke laekaer som separate GET-parametre i udskriv.php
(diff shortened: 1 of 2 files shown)


LuiHedlund02:
The urlencode fixes the reported `&valg=faktura` leakage, but two things
before merge:

1. `$returside` is untrusted GET input and `includes/udskriv.php` later
   emits it raw into JavaScript and href attributes — encoding it for this
   hop doesn't make the decoded value safe. Please restrict it to approved
   same-origin relative paths (and escape per output context where emitted).
2. When neither the caller nor GET supplies it, `urlencode(null)` on the
   `$nomailantal > 0` path raises a PHP 8.1+ deprecation — cast
   `(string)$returside` or default to `''`.

Retest with a return URL containing nested `&`/`?`, with no returside at
all, and with a hostile quote/JS payload.

LuiHedlund02:
Superseded by #534, which includes these commits merged with current master and fixes one issue in the pdftk branch: the `<?= json_encode($returside) ?>` inside the double-quoted string is not evaluated by PHP, so the redirect never ran. See SD-664 / SST-715 for context — thanks for the original fix @momocoder14.

PR #452, opened 2026-08-07
MB1 - buggy kreditor ordre split

SirRolin:
Fix split order multiplying quantity by 1000

usdecimal() strips '.' as a Danish-style thousands separator, meant for parsing user-typed input. Running it on antal/leveret straight from the DB (e.g. "2.000") turned the value into 2000, so splitting a line of 2 left 1999 on the original line (displayed as 1.999) instead of 1.

Use (float) on the DB-fetched values instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed decimal quantities in order lines so values such as 2.000 are preserved correctly instead of being incorrectly converted.
  * Updated order-line quantity and delivery-value handling for more accurate data display and processing.


LuiHedlund02:
The float cast on the `2.000`-style numeric strings is the right fix for the
1000x quantity bug — good find. Before it merges though, this mutation path
needs server-side validation to land with it, since it writes bookkeeping
rows:

1. `linje_id[]`, `MoveItemsTo`, and the quantity arrays reach SQL
   interpolation raw — cast/validate them server-side before any query
   (don't rely on `injecttjek()`).
2. The caps come from hidden POST fields `maxQt[]`/`maxSQt[]`, which a
   client can raise — derive the limits from the freshly fetched DB rows
   inside the transaction and reject out-of-range values.
3. There's no ownership check binding each line ID to the source order and
   the target to the same creditor/context — add one.

Test plan: partial + full moves for `2.000` and fractional quantities,
delivered-line splits, tampered POST IDs/limits, cross-order targets, and
assert nothing in `ordrelinjer`/`batch_kob` can go negative.

SirRolin:
Allowed self-review by Adam

PR #453, opened 2026-08-07
MB 5 rapport visual glitch fix

SirRolin:
- Fixed debitor rapport async loading of OpenPost adding a new div wrapper on each open with extra padding.
- Fixed debitor rapport openpost's table not having padding, as per Adam's request, I made it 12 px instead of 8px.
- Fixed Creditor rapport openpost's table not having padding.
- Added id's to tables of rapport, rapportfunc and showOpenPost for future troubleshooting.

## Summary by CodeRabbit

* **New Features**
  * Open-post reports now load progressively, displaying the layout and table structure before results finish loading.
  * Asynchronous report loading now shows fetch errors when they occur.
* **Bug Fixes**
  * Fixed links for viewing hidden open-post reports.
  * Corrected foreign-currency balance calculations when exchange rates are missing.
  * Prevented report responses from being cached and preserved safeguards against unintended reloads.
* **Improvements**
  * Open-post report tables now use clearer layout identifiers and begin rendering sooner while data is retrieved.


LuiHedlund02:
The padding fix is good, but the PR also broadens the async report shell,
and that part isn't finished:

- The generic async-shell block sits *after* the early GET dispatcher in
  `debitor/rapport.php`, so direct GET links for kontokort/kontosaldo/
  accountChart never reach it — the generalization doesn't apply on the most
  common path.
- Inner open-post navigation in `includes/rapportfunc.php` still omits
  `openpost_content` (the CodeRabbit request). The new `Sec-Fetch-Dest:
  iframe` check only masks it for clients that send Fetch Metadata; others
  recursively render a shell inside the iframe.
- `Cache-Control: no-store` is set after body output has already been
  emitted, so it isn't guaranteed to apply.

**Ask:** either trim this PR to the padding fix alone (fastest), or finish
the shell work: move the block ahead of the dispatcher, add
`openpost_content` to the inner links, set the header before any output —
and then test direct GET entry, POST entry, internal filter/paging links,
and Back navigation for each report type.

SirRolin (commented):
@CodeRabbit review

SirRolin (commented):
@coderabbitai review

PR #457, opened 2026-08-11
Batch expiry fefo fixes

sawaneh1:


LuiHedlund02:
Deferred batch/expiry follow-up; resume after #608 and rebase onto its final implementation. Marked `udskudt` so the new stale workflow cannot close deliberately deferred work. No reviewer request is being added while parked.

sawaneh1:
@LuiHedlund02 master merged in so the branch is mergeable again. varekort.php now uses master's 20260902 layout: the batchExpiryEnabled + batch-group condition wraps the pcSecExpiry box, and the expiry_section marker is kept. Still parked until #608 lands; it overlaps #608 in kreditor/ordre.php and betweenUpdates.php, so I will merge once more after that.

ZaynSaul (dismissed):
Went through all four fixes in the new batch_udlobsdato.txt doc against the actual code, since the PR description itself is empty. All four check out: the three-state save on kreditor/ordre.php and ordreM.php (untouched/cleared/set) matches the read side in openOrderLines.php, the FEFO ordering change in ordrefunc.php correctly falls back to the old FIFO order when due_date is NULL so non-expiry items aren't affected, the column padding fix matches the actual row-rendering markup, and the item-card section is now properly gated on both the tenant setting and the item's batchItem flag.

Also checked the settings call in diverse.php isn't repeating the SD-657 tenant-scoping bug - it isn't, it's intentionally a company-wide toggle same as the packagingModuleEnabled setting right above it, not a per-user preference that got saved globally by mistake.

Only real gap: no automated tests on the purchasing-side writes, and the doc's own FEFO-vs-FIFO repro against ssl12 isn't something I reproduced myself - would want that confirmed live before this goes out given it touches purchase order writes. Small nit: settingsRegistry.php picked up new keywords but no history-block line. Not blocking either of those, but worth a look.

SirRolin (dismissed):

PR #458, opened 2026-08-12
SD-563 Welcome email editor with CSRF, sandboxed preview, literal pla…

sawaneh1:
…ceholder substitution and masked delivery log

The debtor/creditor open posts report had no way to narrow a long list once on screen: no way to show only accounts owing in ">90", no sort by amount, and changing the account range meant going back to the front page. This adds three things to the report itself, with all state carried in the query string so pagination, the BS toggle, udlign links and accountChart links keep it:

Aging filter — each of the five column headers (>90 … 0-8) is a link; clicking one shows only accounts with an amount in that bucket, with a "Ryd filter" link and a "Filter: …" marker.
Amount sort — ▼/▲ on "I alt" (or on the filtered column) sorts by amount desc/asc.
In-report account search — a field in the header accepting a kontonr, a fra:til range or a firm-name pattern (* wildcard), same rules as the front page. kreditor/rapport.php accepts kontonr for this like debitor/rapport.php.
Implementation: the per-account bucket maths moved unchanged out of the display loop into openpost_account_aging(), and the show-rule into openpost_account_visible(), so filter/sort and the rendered page share one implementation — including the SST-717/SST-730 open-balance rules. When a filter or sort is active, a streaming pre-pass aggregates every matching account (from the same grouped superset the count query uses) and pages the resulting ids; without filter/sort the existing queries run untouched. Request values are whitelisted in openpost_report_state() before reaching SQL/URLs. openpost() no longer overwrites dato/konto with the stored DRV row when the request carries konto_fra/kontonr. Texts 5120–5124 added.

Merged with master twice during review; MB-5, SST-672, SST-717/730 and the async-shell changes are all preserved.


LuiHedlund02:
Deferred email-editor work; preserve the September 7 deferral and resolve translation-ID collisions before resuming. Marked `udskudt` so the new stale workflow cannot close deliberately deferred work. No reviewer request is being added while parked.

sawaneh1:
> Deferred email-editor work; preserve the September 7 deferral and resolve translation-ID collisions before resuming. Marked `udskudt` so the new stale workflow cannot close deliberately deferred work. No reviewer request is being added while parked.

@LuiHedlund02 master merged in and the translation-ID collision is resolved: ids 5056–5114 were taken on master, so the welcome-email editor texts moved to 5400–5458 (csv rows + all call sites). Checked against every open PR head, the range is free. PR stays parked (udskudt); no review requested.

ZaynSaul (changes requested):
Went after the three things this PR advertises directly instead of taking them on faith. CSRF: real per-session token via random_bytes, checked with hash_equals plus content-type/origin checks, and a failure actually 403s rather than just logging - holds up. Sandboxed preview: it's a real `sandbox=""` iframe (no allow-scripts, no allow-same-origin) fed via srcdoc, not innerHTML - holds up even against a malicious admin trying to plant something for another admin to preview. Placeholder substitution: fed it an XSS payload and a CRLF header-injection attempt directly against opret_email_render()/opret_email_header_value() - both got neutralized (htmlspecialchars, CR/LF stripped before mail()). So no security blocker from me.

What's stopping me from approving as-is: there's no test coverage anywhere for this - it's a 2000-line feature that sends real mail with rich text and placeholder substitution, and this codebase does have precedent for testing email features (test_kontakt_emails.php), so this shouldn't be the exception. Separately, the PR body doesn't match this PR at all - it's describing a debtor/creditor aging filter, not the email editor. Title and diff are right, just the body needs fixing so it's not confusing for anyone auditing this later.

Once there's some test coverage and the body's fixed I'm good. Non-blocking: no issue with the vendored Quill license (confirmed genuine BSD-3-Clause, version 2.0.3), but there's no build step for the minified JS - worth keeping in mind for future Quill security patches.

sawaneh1:
> Went after the three things this PR advertises directly instead of taking them on faith. CSRF: real per-session token via random_bytes, checked with hash_equals plus content-type/origin checks, and a failure actually 403s rather than just logging - holds up. Sandboxed preview: it's a real `sandbox=""` iframe (no allow-scripts, no allow-same-origin) fed via srcdoc, not innerHTML - holds up even against a malicious admin trying to plant something for another admin to preview. Placeholder substitution: fed it an XSS payload and a CRLF header-injection attempt directly against opret_email_render()/opret_email_header_value() - both got neutralized (htmlspecialchars, CR/LF stripped before mail()). So no security blocker from me.
> 
> What's stopping me from approving as-is: there's no test coverage anywhere for this - it's a 2000-line feature that sends real mail with rich text and placeholder substitution, and this codebase does have precedent for testing email features (test_kontakt_emails.php), so this shouldn't be the exception. Separately, the PR body doesn't match this PR at all - it's describing a debtor/creditor aging filter, not the email editor. Title and diff are right, just the body needs fixing so it's not confusing for anyone auditing this later.
> 
> Once there's some test coverage and the body's fixed I'm good. Non-blocking: no issue with the vendored Quill license (confirmed genuine BSD-3-Clause, version 2.0.3), but there's no build step for the minified JS - worth keeping in mind for future Quill security patches.

@ZaynSaul thanks for checking the security points directly. Both of your points are fixed:

1. Tests: tests/characterization/admin/OpretEmailTest.test.php, 52 tests, no database needed. build/send run in a child process with mail() captured through sendmail_path, so the real message (headers and both MIME parts) is asserted and nothing is sent. Covered: placeholder substitution (XSS, $0/\1/${1} literal, nested {{email}}, unknown placeholders), header injection via the customer name and the sender setting, CRLF in the recipient, the CSRF gate (wrong/missing token, form content types, foreign/look-alike Origin, other port), exactly one message per send, every error_code without anything handed to the MTA, price parsing, log masking, and that error codes and labels exist in tekster.csv.

   The tests found a real bug: the empty_template check in opret_email_send() looked at the wrapped document, which always holds the SALDI header and footer, so an empty template sent an empty mail. It now checks the body's own text.

2. PR body: replaced. It had the SD-140 open-posts description pasted in by mistake. It now describes this PR, lists the tests and has a manual test plan.

On Quill: agreed. The body now notes that it is the upstream minified 2.0.3 release with no build step, and that a security update means replacing quill.min.js and quill.snow.css.

PR #464, opened 2026-08-13
sd 140 openpost aging filter sort search

sawaneh1:

The debtor/creditor open posts report had no way to narrow a long list once on screen: no way to show only accounts owing in ">90", no sort by amount, and changing the account range meant going back to the front page. This adds three things to the report itself, with all state carried in the query string so pagination, the BS toggle, udlign links and accountChart links keep it:

Aging filter — each of the five column headers (>90 … 0-8) is a link; clicking one shows only accounts with an amount in that bucket, with a "Ryd filter" link and a "Filter: …" marker.
Amount sort — ▼/▲ on "I alt" (or on the filtered column) sorts by amount desc/asc.
In-report account search — a field in the header accepting a kontonr, a fra:til range or a firm-name pattern (* wildcard), same rules as the front page. kreditor/rapport.php accepts kontonr for this like debitor/rapport.php.
Implementation: the per-account bucket maths moved unchanged out of the display loop into openpost_account_aging(), and the show-rule into openpost_account_visible(), so filter/sort and the rendered page share one implementation — including the SST-717/SST-730 open-balance rules. When a filter or sort is active, a streaming pre-pass aggregates every matching account (from the same grouped superset the count query uses) and pages the resulting ids; without filter/sort the existing queries run untouched. Request values are whitelisted in openpost_report_state() before reaching SQL/URLs. openpost() no longer overwrites dato/konto with the stored DRV row when the request carries konto_fra/kontonr. Texts 5120–5124 added.

Merged with master twice during review; MB-5, SST-672, SST-717/730 and the async-shell changes are all preserved.

## Summary by CodeRabbit

* **New Features**
  * Added account search to the open posts report, supporting individual accounts, ranges, and firm-name patterns.
  * Added aging-bucket filters and amount sorting.
  * Added clickable report headers and an improved search row.
  * Preserved selected filters and report settings when paging, switching views, or performing actions.

* **Bug Fixes**
  * Date range and return-page filters now remain active during account searches.
  * Report results and page counts remain accurate when filtering or sorting.


LuiHedlund02:
Pre-deploy review. The filter/sort/pre-pass core is sound and correctly escaped, but three real defects need fixing before merge (the rebase itself is only tekster.csv, ids 5120-5124 are free):

1. **`$currentdate` is gone.** The date block was moved up in `includes/reportFunc/showOpenPosts.php` but `$currentdate=date("Y-m-d");` was dropped and nothing else assigns it (used at :390, :495, :502 on the PR head). With blank date filters `$todate` becomes null: overdue amounts vanish (an unpaid invoice due yesterday shows no overdue), and dates ending today take the historical branch, whose net-balance `HAVING` hides accounts with unmatched +100/-100 posts. Re-add the assignment before the block.
2. **Raw search term reaches the reminder query.** `debitor/rapport.php:289` assigns `openpost_kontonr_range($_GET['kontonr'])` after the escaping stage; `openpost_account_filter()` escapes its own query but the reminder query further down interpolates `$konto_fra`/`$tmp1`/`$tmp2` directly. Searching `O'Brien` breaks it; `x') OR 1=1 -- -` matches every order. Escape at that query, or parse the range inside the :212-222 block with the neighbours.
3. **DRV row written before the range is parsed.** `rapport.php:280` updates the remembered range (`box4`/`box5`) using `$konto_fra`/`$konto_til`, which line :289 only sets afterwards. Submitting the new search box blanks the remembered range (PHP 8 warning too), and the next POST from that page ("Ryk alle") reloads the DRV row and renders **all** debtors instead of the searched account. Move :289 above :280.

Non-blocking: the mode-switch options at `rapportfunc.php:235-239` print `$konto_fra` raw into an attribute and drop the aging state; "▼" sorts creditors smallest-debt-first (sort by `abs()`); ~20 merge commits on the branch, squash-merge please.

SirRolin (approved):

PR #465, opened 2026-08-13
Bilagsmatch: pinned document preview

ZaynSaul:
## Summary
- Adds a click-to-open, persistent document preview to the Bilagsmatch dialog (kravsspecifikation popup fast.pdf): draggable/resizable window, native PDF rendering (fit-to-width, page nav, zoom via the browser's own viewer), custom zoom for images, and clean loading/error/unsupported-format states.
- Retires the old hover-only preview - the pinned window is now the single, consistent way to view an attachment, which also removed a visual conflict where the hover box could compete with the pinned window near the cursor.
- Polish from live testing: centered default position, redesigned the loading/error/unsupported states (light status card + icon/spinner instead of a flat dark canvas), and gave the attachment icon a button-like chip so it reads as clickable without relying only on a tooltip.
- Fixes a stale-translation gap: findtekst() prefers an existing `tekster` DB row over `tekster.csv`, so a `betweenUpdates.php` migration clears the old tekst_id 5068 row so the updated tooltip text actually takes effect.
- Merges in latest `master`, notably reconciling with #454 (ambiguous `norm_amount` column fix) - this branch had its own independent, incomplete attempt at the same fix; upstream's version is the one now in place.

## Test plan
- [x] Verified live (headless browser) against a real running instance: open/replace/close the pinned window, hover-preview suppression, drag/resize clamped to viewport, row selection/scroll unaffected while open, Esc/close-button focus return.
- [x] Verified PDF (native viewer, fit-to-width, page nav), image (zoom controls), unsupported format (message + working download link), and missing-file (error + retry) against real local test fixtures.
- [x] Verified translations across Danish/English/Norwegian.
- [x] Confirmed the merged `fetchbilagsmatch.php` query runs without the "ambiguous column" error and returns scored matches against local test data.
- [x] Known gap, not addressed here: document access has no rights check (NFR-6) - pre-existing, app-wide (also in includes/documents.php, docPool.php), not introduced by this change.

## Summary by CodeRabbit

* **New Features**
  * Added configurable document matching for date range, amount tolerance, and posted documents.
  * Added live result counts, loading feedback, match badges, and confirmation for uncertain matches.
  * Added persistent, draggable attachment previews with zoom, resizing, downloads, and new-tab opening.
* **Bug Fixes**
  * Improved matching reliability and prevented outdated results from replacing newer searches.
  * Standardized currency and amount handling across imported and edited documents.
  * Improved selection controls, empty-result handling, and confirmation states.


ZaynSaul:
@LuiHedlund02 rebased and pushed — ready for your review.

Heads up on scope: this wasn't a mechanical rebase. The branch was 51 commits behind master and had gone through 6 "merge master in" cycles over its life, so a lot of its own history (the original scoring-engine rewrite, PopupManager work, REST API test suite) had already landed on master independently through other PRs. Untangling that dropped it from 17 commits down to 3 - final diff is confirmed identical to the original PR's file list (`gh pr diff --name-only`), nothing added or dropped.

One real conflict, not just noise: this PR's new text strings collided with tekst_ids master had independently claimed in the meantime (both sides picked 5056+ unaware of each other). Renumbered the 19 new strings (5056-5074 → 5059-5077) consistently across tekster.csv, the findtekst() calls in bilagsmatch.php, and one migration statement in betweenUpdates.php that referenced a specific id by number.

Verified live against a real tenant DB post-rebase (not just diff/lint):
- Popup renders with all renumbered strings showing correct Danish text, zero PHP warnings on page load.
- betweenUpdates.php's Bilagsmatch migration (pool_files.norm_amount, pg_trgm) confirmed already applied cleanly.
- Hit fetchbilagsmatch.php directly with a synthetic pool_files row - scoring engine correctly matched it against a real kassekladde line (amount_score 40, date_score 25, score 65, high precision), then cleaned up the test row.
- The actual pinned-preview widget (drag/resize/zoom/PDF) is untouched - confirmed byte-for-byte identical to the pre-rebase branch, so whatever live-testing already happened on that (per the existing commit history) still holds.

Given this touched kassekladde + a migration (both on your own escalation list), flagging for extra scrutiny rather than treating it as routine.

SirRolin (approved):

PR #466, opened 2026-08-13
SST-724 Serial number overview: only show salgslinje_id = 0 as availa…

sawaneh1:
…ble, not credited history rows

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the “not sold” filter to exclude historical return records while retaining newly created return serial records.
  * Updated related documentation to clarify the filter’s behavior.

### Description

**What are the changes about?**

One-line change in `lager/lister/serialnumber.php`: the "Vis kun serienumre der ikke er solgt" filter now uses `sn.salgslinje_id = 0` instead of `<= 0`.

`serienr.salgslinje_id` has three states: `0` = in stock, `> 0` = sold on that sales line, `< 0` = history marker for a sale that was later credited/returned (`debitor/serienummer.php` flips the original row to `-linje_id`, and `krediter()` in `includes/ordrefunc.php` inserts a fresh row with `salgslinje_id = 0` for the returned unit).

Before: `<= 0` matched both the negative history row *and* the fresh row, so a serial that was sold and then credited appeared **twice** in the "not sold" list. Worse, if that serial was sold again, the fresh row got a positive id but the old negative row still matched, so a serial that was currently sold kept showing as available.

After: only rows with `salgslinje_id = 0` count as available — one line per unit actually in stock. Credited returns still appear (via the fresh row), history rows do not.

**How I verified:** [confirm] On a test tenant, sold a serial-tracked item, credited the invoice with the serial returned, then re-sold it. Before the fix the serial number overview with "ikke solgt" ticked listed the serial twice after the credit and still listed it after the resale; after the fix it appears once after the credit and disappears after the resale. Unticking the filter shows all rows unchanged.

### DIFF
--- a/lager/lister/serialnumber.php
+++ b/lager/lister/serialnumber.php
@@ -29,6 +29,7 @@
 // 20260710 MJ Use COALESCE(v.varenr, kl.varenr, sl.varenr) so serienr records with stale/missing vare_id still appear and are searchable by varenr.
 // 20260710 MJ ABS(sn.kobslinje_id) i JOIN så negative kobslinje_id (retur til leverandør) også viser indkøbsordren.
 // 20260710 MJ Ekstra COALESCE-fallbacks via ordrelinjer.vare_id→varer og batch_kob/batch_salg.vare_id→varer så serienr med tom/manglende ordrelinjer.varenr stadig søges.
+// 20260813 Sawaneh - "Not sold" filter: sn.salgslinje_id = 0 instead of <= 0, so negative history rows (credited sales) are no longer shown as available. Credited return serials still appear via the fresh row with salgslinje_id = 0 from krediter().
 
 @session_start();
 $s_id = session_id();
@@ -240,7 +241,7 @@ $filters[] = array(
         array(
             "name" => "Vis kun serienumre der ikke er solgt",
             "checked" => "",
-            "sqlOn" => "sn.salgslinje_id <= 0",
+            "sqlOn" => "sn.salgslinje_id = 0",
             "sqlOff" => "",
         )
     )


ZaynSaul:
The change itself looks right on inspection: switching the "not sold" filter from sn.salgslinje_id <= 0 to = 0 means negative-valued history rows (credited sales) no longer count as available, while the fresh row krediter() inserts with salgslinje_id = 0 for a credited return still does. Consistent with the krediter() serial-row convention in #467.

But the PR description is still the unfilled template — no explanation of the change, and no "how I verified this." Could you fill that in before I approve? Specifically: what did this look like before/after for a serial number that was sold then credited/returned — did it show up twice, or not show as available at all, prior to this fix?

LuiHedlund02:
Review evidence for the bounced items (verified against current master, 2026-08-28):

The filter change `sn.salgslinje_id <= 0` → `= 0` is correct. The `<= 0` form (introduced 20260608 to show return serials) also matches **negative** `salgslinje_id` rows, which are credited-sale history rows — that's exactly SST-724's complaint (sold serials 382030/911000 showing as available). The header comment's assumption holds on master: `krediter()` inserts a fresh replacement row with `salgslinje_id = '0'` for credited serials — see `includes/ordrefunc.php:342` and `:842` — so credited returns still appear as available through that row, independent of the `batch_kob_id` fix in #467.

@sawaneh1 add the verification line to the description and this is ready for approval.

ZaynSaul (approved):
Clean, focused fix. Single-line SQL condition change (`<= 0` → `= 0`), no injection risk (literal, no user input), matches title exactly. Description has a real root-cause explanation (the double-count/still-shows-as-available bug from history rows) and a concrete before/after verification narrative (sell → credit → resell cycle). CodeRabbit approved with zero findings. Lints clean, mergeable.

LuiHedlund02 (approved):
In-house sign-off per master ruleset — reviewed and approved by Saul (small-fix lane).

PR #468, opened 2026-08-14
Fixed: Kassekladde position control handle not showing

Loe4g:
Changes
Render a drag handle unconditionally for every row.
Add pos as a valid/default sort mode and fetch rows with ORDER BY pos, bilag, transdate, id.
Simplify the reorder request, and reload with kksort=pos.

Clear all output buffers before returning the 204 reorder response.
Security

 A pre-existing session/ownership check for kladde_id remains a follow-up item.

Testing
Verify every row has a visible/aligned drag handle.
Verify drag order updates and persists after reload.
Verify Bilag/Dato/Beløb sorting still overrides the custom order.
Verify no console errors and expected pos updates in error_log.

## Summary by CodeRabbit

* **New Features**
  * Added position-based sorting for saved cash-journal views with consistent tie-breaking.
  * Journal rows consistently display draggable position handles.
  * VAT selections preserve saved codes and account defaults, including shortcut accounts.
  * Posted entries use stored VAT codes in grids and displays.

* **Bug Fixes**
  * Improved VAT-exempt handling and validation, including confirmation for one-sided selections.
  * Invalid blank VAT submissions can be restored from account defaults.
  * Improved drag-and-drop recovery after failed saves and while edits are pending.
  * Restored safer field focus and scrolling after edits.
  * Removed journal headers from print-only output.


Loe4g:
This PR has been updated to work with current kassekladde.php @SirRolin

SirRolin (dismissed):

ZaynSaul:
Holding this rather than merging as a freebie — dug into the unresolved CodeRabbit finding on drag-handle-ajax-call.php and it's real, not a false positive.

Traced the full flow:
- onStart (drag begins): if unsavedChanges is already true, it auto-saves the form first, resets unsavedChanges = false, then replays the drag.
- onEnd (drag completes): POSTs the new row order. On success it navigates to a new URL; on failure it calls window.location.reload(). Neither path checks unsavedChanges.

So the gap: if a user starts typing in another field while the reorder POST is in flight (a normal network round-trip, not instant), unsavedChanges flips back to true via the input listener, but the success/failure handlers don't check it before navigating away or reloading — those edits get silently discarded. Narrow window, but a genuine data-loss path, and this also touches finans/kassekladde.php with real new behavior (not a warning-suppression pass), which is on the escalation list.

Could you have the success/catch handlers check unsavedChanges before navigating — e.g. skip the auto-navigate/reload and instead prompt the user (or just refresh the position numbers in place) if changes came in during the request? Holding off on merging until this is addressed.

Loe4g:
Fixed the race — now checking unsavedChanges before navigating/reloading, so in-flight edits from the user aren't discarded. Pushed.

ZaynSaul:
Confirmed fixed — traced through the new code:

- Success path: if unsavedChanges is true when the reorder response comes back, it returns without navigating (edits preserved). Otherwise navigates with kksort=pos as before.
- Failure path: alerts the user, and only reloads if !unsavedChanges — deliberately skips the reload when there are unsaved edits, accepting a stale drag-order display over discarding the user's typed changes. Reasonable tradeoff.
- The success-signal change (JSON body's data.success -> res.ok against the new HTTP 204 response) is consistent with moveButton.php's change to http_response_code(204) with no body.

Approved.

ZaynSaul (approved):
Race condition fix verified — see comment above. Approved.

LuiHedlund02 (approved):

PR #478, opened 2026-08-18
Stripe subscriptions: native signed-link flow, webhook (record-only), settings + mailer token

LuiHedlund02:
Native Stripe subscription support inside the monolith. Dormant by default everywhere it deploys: activation requires a per-install tenant pin (gitignored dotfile/env), API keys in settings, and an explicit **Aktiv** switch (default off). With test-mode keys the key-prefix guard makes live charges impossible.

## What it does
- Invoice mails can carry `$abonnementslink` — a signed URL (HMAC over the order id) derived at send time from that order's lines. Customers confirm on a Danish summary page and pay via Stripe Checkout (card + Link, 25% moms via tax rate).
- `api/stripe/webhook.php` receives Stripe events (signature-verified, DB-deduplicated) and runs **record-only**: payments are recorded in `stripe_events`, the invoice PDF lands in the bilagspulje, and the bookkeeper is alerted to book manually. Automatic booking exists but is gated behind `import_enabled` + boot asserts and stays off until the posting-atomicity work is reviewed.
- Admin: new "Stripe abonnement" sektion under Indstillinger -> Diverse (secrets write-only), CLI seed script `admin/stripe_setup.php`, CLI link printer `tools/stripe_print_link.php`.
- Included hardening: the mail template token resolvers (sendMail/oldSendMail, subject+body) now gate tokens through a column allowlist (closes an SQL interpolation surface); plain-text MIME parts no longer carry raw anchor markup; `oldSendMail`'s `$betalingslink` fatal (called a non-existent function) fixed.

## Blast radius
1,992+ lines across 24 files, but only 6 existing files touched, all additive/guarded: `admin/opret.php`, `includes/betweenUpdates.php` (idempotent DDL, deliberately NOT in the already-run opdat_4.3 gate), `systemdata/diverse.php`, `systemdata/settingsRegistry.php`, `includes/formFuncIncludes/sendMail.php`, `oldSendMail.php`, `api/send_invoice_pdf.php`. `ordrer.nextfakt` is never read or written by any of this code (enforced by design — recurring-order handling stays manual).

## Evidence
- Full live end-to-end on ssl12/lui against the Danosoft Stripe sandbox: signed link -> confirm -> Checkout payment -> Stripe-delivered webhooks (session completed + invoice.paid) -> recorded + PDF + alerts. Already-subscribed guard verified live.
- Release gate (testlab) on the branch: all findings identical to the master baseline — zero new findings introduced; migration ran green on all four golden datasets.
- Mailer verified against mailpit with the real send_mails(): HTML anchor + bare-URL text part, allowlist accept/reject, empty-render for non-subscription orders.

Design contract: `doc/stripe/INTERFACE_CONTRACT.md` (in this PR) — settings schema, DDL, link signature, webhook identity resolution, deploy guards.


.gitignore
```diff
@@ -89,3 +89,6 @@ index/ch/
 # Live Service
 /index/redirect.php
 email.php
+
+# Stripe integration: per-install tenant pin, never committed (doc/stripe/INTERFACE_CONTRACT.md)
+includes/stripeIncludes/.ht_stripe_db
```
SirRolin: Check if actually needed, we have a line higher up ".ht*" which should filter it out.

SirRolin (approved):

PR #479, opened 2026-08-18
Sst 730

sawaneh1:


ZaynSaul (approved):
Sanity-read only, per Lui's note that this was already verified end-to-end on real data on the 21st.

- rapport.php: GET-param fallback for dato_fra/dato_til/konto_fra/konto_til on openpost requests, properly escaped via db_escape_string, only applied when not already set from POST — contained, matches the stated "filters dropped on GET nav" bug.
- showOpenPosts.php: the $visY/$visKontrol split (open-only vs. full historical sums, gated by $vis_alle) is threaded consistently through account visibility, running totals, and the settle-all eligibility check. The settle-all expansion specifically targets the described case (whole-account balance ~0 despite a nonzero open remainder, e.g. currency-difference-settled accounts) without weakening the existing $kun_debet/$kun_kredit alignment logic.

php -l clean on both files, 3 commits behind master, no overlap with either touched file. Approved.

LuiHedlund02 (approved):

PR #500, opened 2026-08-21
Debtor card: keep stored emails when a save posts without the kontakt…

sawaneh1:
…_email fields (JOB-083)


sawaneh1:
The debitorkort save handler rewrote kontakt_emails and adresser.email on every save, so a POST without the kontakt_email fields (e.g. a stale pre-deploy tab) deleted the customer's primary and extra emails 
The rewrite and sync now only run when the form actually posted the email fields; `email` is dropped from the main adresser UPDATE since the sync afterwards was already the effective writer, so normal saves are unchanged.

SirRolin (approved):

PR #502, opened 2026-08-24
Fix MobilePay silent auth failures, add QR regeneration, persist paym…

momocoder14:
- Surface Vipps accesstoken/get failures as a clear error instead of hanging on "Afventer QR scan" or continuing with a null token (mobilepay.php, mobilepayListen.php, sys_div_func.php QR setup)
- Add a "Regenerer QR-kode" button so a stale till QR can be refreshed without manual DB edits (sys_div_func.php + regenerate_mobilepay_qr.php)
- Persist the Vipps payment reference into pos_betalinger.receipt_id for completed MobilePay payments: pass the reference through the success redirect, rename afslut()'s unused $line parameter to $receipt_id so it actually reaches settlePOS.php, and fix settlePOS.php's (int) cast which zeroed out the varchar column


LuiHedlund02:
Draft MobilePay work remains deferred; renumber the colliding IDs and finish its verification before requesting review. Marked `udskudt` so the new stale workflow cannot close deliberately deferred work. No reviewer request is being added while parked.

ZaynSaul (changes requested):
The actual fixes here are good - traced all three through. The silent-auth-failure fix is real: before, none of the three files checked the HTTP status or whether access_token actually came back, so a bad/expired credential fell through with an invalid token and produced confusing errors two calls later instead of failing clearly - now it checks status+token and surfaces findtekst 5057. The receipt_id fix is also real: pos_betalinger.receipt_id is varchar(75), but settlePOS.php was casting it to (int), silently zeroing any real Vipps reference like `ordreid-123-1234` - now it's escaped as a string instead, correctly quoted on both the insert and update. No injection risk anywhere I could find, including the new QR endpoint.

Why I can't approve as-is: this PR conflicts with master right now (mergeable: CONFLICTING) because it adds tekster.csv ids 5057/5058, and master has since given those same two ids to two unrelated strings from other merged PRs. That needs a rebase and renumbering before this can go in - not optional, GitHub is refusing the merge itself. On top of that there's no test or live-verification evidence anywhere in the PR for a payment-persistence change, which given this is money-adjacent I'd want before it merges.

Non-blocking: none of the touched files got a history-block line even though the pattern is followed elsewhere in this codebase, and the QR-regen endpoint has no explicit permission check beyond session/tenant resolution - same posture as its sibling webhook files though, so not a regression, just worth someone's attention eventually.

PR #508, opened 2026-08-25
Fix: MB-18 visual label editor silently destroys templates it can't fully r…

ZaynSaul:
## Summary
Changing any label setting via the visual editor could overwrite/destroy a customer's
entire existing label configuration, instead of applying only the specific change made.

## Root Cause
The visual label editor doesn't patch individual settings — every "Save" click rebuilds the
label's entire HTML from scratch (`generateLabelTemplate()`) using a narrow, fixed set of
fields (columns, rows, margins, a handful of checkboxes, up to 5 custom text lines). Its
counterpart, `parseLabelTemplate()`, can only recognize that exact same narrow shape when
reading a template back in. Anything outside it — custom CSS like `transform: rotate()`,
non-default dimensions, extra markup — is invisible to the parser, and is silently thrown
away the moment the visual editor regenerates and saves.

This wasn't a rare edge case: every label template Saldi ships (Brother 22606, Dymo 11354,
both "mit salg" variants) uses CSS the parser doesn't understand. Opening any of those in
the visual editor and saving any single setting — even something as trivial as text length —
would silently wipe out the real layout and replace it with generic filler.

- Added `labelTemplateEditableVisually($labelText)` — runs the template through
  `parseLabelTemplate()` → `generateLabelTemplate()` and checks whether that reproduces the
  original exactly. If it doesn't, the template has content the visual editor can't safely
  represent.
- The editor page (`labels()`) now forces raw-HTML mode, with an explanation, for any label
  that fails this check, instead of offering the destructive visual editor; "Switch to
  Visual Editor" is hidden for unsafe labels.
- The `$saveLabel` POST handler (`systemdata/diverse.php`) enforces the same check
  server-side — a save is refused, not silently destructive, whenever the label's current
  template isn't visual-editor-safe. This is the authoritative guard; the UI change alone
  couldn't be trusted to block every path.

## Files Changed
- systemdata/sys_div_func.php — new `labelTemplateEditableVisually()` function; `labels()`
  now forces raw-HTML mode and hides "Switch to Visual Editor" for unsafe labels
- systemdata/diverse.php — `$saveLabel` POST handler now checks the guard before
  regenerating/saving
- tests/characterization/systemdata/LabelTemplateEditableVisuallyCharacterizationTest.test.php
  (new) — 8 tests

## How to Test
1. Open each of the 4 shipped import templates (Brother 22606, Dymo 11354, both "mit salg"
   variants) in the label editor — verify each is forced into raw-HTML mode with an
   explanation, and "Switch to Visual Editor" is not offered.
2. Attempt to save via the visual editor path against one of those unsafe templates directly
   (bypassing the UI, e.g. via the POST handler) — verify the save is refused server-side and
   the template is preserved byte-for-byte.
3. Create/open a genuinely visual-editor-built label (safe template) and change a setting —
   verify it still saves normally, confirming the guard doesn't over-block legitimate use.
4. Open an empty label — verify it's correctly flagged safe.
5. Run `LabelTemplateEditableVisuallyCharacterizationTest.test.php` and confirm all 8 cases
   pass.
6. Reproduce the original destructive scenario against the pre-fix code (stash the fix,
   seed a real imported Brother template, simulate a "change one field" save) and confirm it
   destroys the label (`rotate` gone, real 50mm width replaced by generic 38.1mm) — proving
   the test suite actually pins this bug.
7. Restore the fix and re-run the same scenario — confirm the save is refused and the
   template is preserved exactly.

## Acceptance Criteria
|---|---|---|
| 1 | All 4 shipped import templates | Correctly flagged unsafe for visual editing |
| 2 | Visual-editor-native label | Correctly flagged safe |
| 3 | Empty label | Correctly flagged safe |
| 4 | Save attempt against an unsafe template | Refused server-side, zero data loss, template preserved byte-for-byte |
| 5 | Save attempt against a safe (visual-editor-native) template | Still saves correctly |

## Verification
- Stashed the fix, seeded a real imported Brother template, simulated a "change one field"
  save — confirmed it genuinely destroyed the label (`rotate` gone, real 50mm width replaced
  by generic 38.1mm).
- Restored the fix, reran the same scenario — refused, template preserved byte-for-byte.
- Confirmed a genuinely visual-editor-built label still saves normally (guard doesn't
  over-block).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented visual label editing from overwriting templates that cannot be reproduced accurately.
  * Automatically switches unsupported templates to raw HTML mode and displays a warning when saving is refused.
  * Improved handling of imported labels, legacy templates, custom text, and decimal values.
  * Preserved editable templates and valid visual-editor changes during saving.

* **Tests**
  * Added coverage for imported, empty, and visually editable label templates, including varied formatting and decimal separators.


SirRolin (changes requested):
Performed test:
Opened raw html editor.
Added a line.
Saved.
It correctly blocked me from accessing visual editor.
Deleted everything in the template.
Saved. -> it regenerated the default template.
It incorrectly blocks me from accessing the visual editor.

ZaynSaul:
@SirRolin Fixed in 8065a5f9 — thanks for the repro steps, they pinpointed it exactly.

Root cause: in `sys_div_func.php`'s `labels()`, when a `'Standard'` label's stored text is empty, a hardcoded placeholder template is substituted in for display in the raw-HTML textarea (it still uses the older `$beskrivelse`/`$pris` variable names, predating `generateLabelTemplate()`'s `$minbeskrivelse`/`$minpris`). The new MB-18 guard was running `labelTemplateEditableVisually()` on that placeholder instead of on the actual stored (empty) text, so it always evaluated to "not visually editable" for an emptied-out label — even though there's nothing there to lose.

Fix captures the stored text before the placeholder substitution and guards on that instead, so an emptied `'Standard'` label correctly regains access to the visual editor. Added a regression test (`testTheEmptyStandardDisplayPlaceholderIsNotItselfVisuallyEditable`) pinning the placeholder-vs-stored-text distinction that caused this.

Also swapped the hardcoded Postgres credentials in `LabelTemplateEditableVisuallyCharacterizationTest` for the gitignored `connect.php` convention (935dd29f) — same pattern as the fix on #507.

Could you re-test the same repro when you get a chance?

ZaynSaul:
@SirRolin following up — could you re-test when you get a chance? This is the top of my queue for Monday's deadline (MB-17/MB-18 both hang on it).

To directly re-verify your repro: raw HTML editor → delete everything → save. It now correctly returns to the visual editor instead of staying locked in raw mode. Added a regression test pinning exactly that (`testTheEmptyStandardDisplayPlaceholderIsNotItselfVisuallyEditable`), plus re-ran the full suite live against real DB rows on both the pre-fix and post-fix code to confirm the before/after behavior.

Also cleaned up a stray `if_isset()` warning in the test helper itself (same calling-convention bug as a couple other tickets this week) — suite runs with zero warnings now.

No other changes since your review — just the targeted fix for what you found.

ZaynSaul:
Status for today's deadline:

- **Rebase/conflicts**: merged current master in (fdcaa1ef, pushed this morning) — GitHub now shows `MERGEABLE`, no conflicts.
- **CodeRabbit's two new findings** (on the merge push) — both addressed in 01f83910: `tearDownAfterClass()` isn't guaranteed to run after a `setUpBeforeClass()` skip, so the cwd restore now happens before `markTestSkipped()` instead of relying on teardown; and the test's cleanup delete is now scoped to `account_id = '0'/null` (matching what `saveLabelText()` actually writes) so a same-named account-specific label in the cloned tenant can't be collateral damage.
- Full characterization suite (9 tests) still green locally after both changes.

@SirRolin — this is now blocking MB-17/MB-18 from tonight's release (freeze is tomorrow 18:00). The fix for your repro (delete a Standard label's raw HTML, save, get stuck out of the visual editor) has been up since Aug 26 (8065a5f9) with a regression test pinning exactly that case, and I followed up twice since with no response. Could you re-run your repro today so this can go out, or let me know if something's still off?

LuiHedlund02 (commented):
Reviewed against `fdcaa1e`+`01f8391`, the conventions in `doc/ai/`, and MB-18.

The core idea is right — a parse→generate round-trip is the correct shape for this guard, and the ticket describes a genuinely destructive bug worth stopping. The problem is that the round trip is **not idempotent on the visual editor's own output**, so the guard both over-blocks untouched legacy data and can lock a user out of a label the editor itself just wrote. Since `$forcedRawHTML` sets raw mode *and* hides "Skift til Visuel Editor", there is no way back except hand-editing the HTML into a shape the parser happens to reproduce.

## Blocking

**1. `systemdata/sys_div_func.php:2485` — every upgraded install's Standard label gets locked out, behind a message that isn't true.**

`parseLabelTemplate()` reads `$beskrivelse`/`$pris`, but `generateLabelTemplate()` always writes `$minbeskrivelse`/`$minpris` (deliberate, per the 20260824 history entry). A template using the item-own variables therefore can never round-trip.

`includes/opdat_4.0.php:806-815` copies each install's `grupper.box1` into `labels.Standard` on upgrade — and that legacy default is byte-identical to `generateLabelTemplate()` output *except* for those two variable names. This PR's own `testTheEmptyStandardDisplayPlaceholderIsNotItselfVisuallyEditable` proves it: the placeholder it asserts as unsafe differs from generator output only in `$beskrivelse` / `Pris $pris`.

Net effect: on every existing installation that never re-saved through the visual editor, the Standard item label now renders forced-raw with a red banner saying it "indeholder formatering, som den visuelle editor ikke forstår". Nothing about it is unrepresentable. Same for any customer deliberately using `$pris`/`$beskrivelse`, which the raw-HTML help text (tekst 5057) explicitly documents as supported.

Suggested fix: canonicalise the two known-equivalent renames before the equality comparison. Templates that differ in more than the rename stay blocked, correctly.

**2. `systemdata/sys_div_func.php:2485` — the visual editor can produce a template the guard then refuses, locking the label out on the next page load.**

Set Varenummer to 14px and Mærke to 16px. With both `show_varenr` and `show_varemrk` on and sizes differing, `generateLabelTemplate()` emits both spans on one line. The varemrk regex

```
/<span[^>]*font-size:\s*([0-9.]+)px[^>]*>.*?\$varemrk.*?<\/span>/i
```

matches leftmost, so the *first* `<span … 14px>` opens the match and `.*?` runs across into the second span. It captures **14**, not 16. Round trip differs → flagged unsafe → the label is forced to raw HTML on the next load, with no route back to the visual editor.

Same one-way trap from several other inputs the editor's own UI accepts:
- custom text containing `Pris`/`pris`, `$`, or `<img` — the parser's custom-line filter drops those lines entirely (`Pris pr. stk`, `Kun $99`);
- more than 5 custom text lines — the parser stops at 5;
- a Danish decimal comma in the free-text dimension fields — `3,5` yields `margin-left: 3,5mm`, which `/margin-left:\s*([0-9.]+)mm/` cannot read, so the round trip substitutes a different value.

In each case: the save is permitted (the *old* template was safe), and the label is locked out on the next render. The guard should be evaluated against what the editor would write for the current form state, not against a parser whose gaps the editor can reach.

**3. Scope — this mitigates MB-18 rather than fixing it.**

The ticket says changing a setting should "only apply the specific change made". This PR stops the destruction by removing the visual editor instead. All four shipped templates (Brother 22606 ×2, Dymo 11354 ×2) fail the check, so customers on a shipped template now have no visual editor at all and must hand-edit HTML to change a font size. That's a defensible stop-the-bleeding trade, but it is a product decision — flagging it so MB-18 doesn't close as done on it.

**4. No escape hatch, and the warning doesn't offer one.**

Once locked out, the only way back is to blank the raw HTML and save (empty → safe again). The message doesn't say so. Either mention it, or keep "Skift til Visuel Editor" behind a confirm ("dette sletter den formatering, editoren ikke kan vise").

**5. `systemdata/diverse.php:1253` — the server-side refusal is a silent no-op.**

The `if (!labelTemplateEditableVisually(...)) { }` branch has an empty body; execution falls through to `labels($valg)`, which re-renders the **stored** template. The user's submitted values are discarded with no "save refused" message — only the generic banner. This is reachable whenever the template changed between page render and submit (second tab, another admin, back-button re-post), which is exactly the race this guard exists to catch. Invert the condition (`if (editable) { …save… }`) and set an explicit error flag.

**6. `tests/characterization/systemdata/LabelTemplateEditableVisuallyCharacterizationTest.test.php:249` — the guard described as authoritative has no coverage.**

`attemptGuardedVisualSave()` re-implements the `diverse.php` `$saveLabel` branch, including its own `if (!labelTemplateEditableVisually(...)) return;`. So `testSavingOneSettingOnAnImportedLabelIsRefusedAndNothingIsLost` would still pass if the guard were deleted from `systemdata/diverse.php` entirely — it only exercises the helper. The PR description calls that check "the authoritative guard a form submit cannot bypass"; nothing pins it. `labels()`'s forced-raw path is untested too.

## Should fix

- **`systemdata/sys_div_func.php:2531`** — the new warning is a hardcoded Danish literal. Every other user-facing string in `labels()` goes through `findtekst('id|text', $sprog_id)`, including the line directly above it (tekst 5057). A non-Danish user gets Danish for the one message explaining why their editor is disabled. Needs its own tekst id.
- **`systemdata/diverse.php:1252`** — the new `else` wrapper leaves ~30 lines at the original indent level. `doc/ai/feedback_whitespace_and_indentation.md` asks for indentation to be corrected in the block being edited. Inverting the condition per #5 removes this entirely.
- **`systemdata/sys_div_func.php:2896`** — `preg_replace('/\s+/', '', $s)` strips whitespace *inside* content, not just between tokens, so two templates differing only in internal spacing compare equal and a "safe" save silently rewrites the stored text. Collapsing runs to a single space matches the docblock's stated contract better than deleting them.
- Docblock on `labelTemplateEditableVisually()` has no `@param`/`@return` (`doc/ai/feedback_documentation_style.md`), and `} # endfunc labels` now trails the new function — worth correcting while it's being touched.

## Status of earlier feedback

- @SirRolin's finding (emptying the template left you locked out) is addressed by `8065a5f` and pinned by `testTheEmptyStandardDisplayPlaceholderIsNotItselfVisuallyEditable` — worth a re-test, since the CHANGES_REQUESTED is still the merge gate.
- Both CodeRabbit nits (cleanup scoping, cwd restore on skip) are fixed in `01f8391`. CI is green.

Findings 1 and 2 are from static reading of `parseLabelTemplate()`/`generateLabelTemplate()` and the 4.0 migration, not from an executed run — running the suite against a seeded `saldi_chartest` tenant would confirm both.

ZaynSaul:
Thanks for this - both blocking findings checked out, confirmed by actually running the round-trip against real input, not just reading the code. Pushed 5626646.

**Blocking**

1. Fixed. Confirmed empirically: the exact legacy placeholder (opdat_4.0.php's migrated `$beskrivelse`/`$pris` shape) came back `editable? false` before, `true` after. `labelTemplateEditableVisually()` now canonicalizes the old names to `$minbeskrivelse`/`$minpris` before comparing. Updated the existing regression test (`testTheEmptyStandardDisplayPlaceholderIsNotItselfVisuallyEditable` → `...IsAlsoVisuallyEditableOnItsOwnMerits`) since the correct expectation flipped.

2. Fixed. Reproduced your exact scenario (varenr 14px / varemrk 16px) - the varemrk regex was capturing 14, confirmed. Root cause was `.*?` crossing the `</span>` boundary into the other span; switched to `[^<]*?` on all four per-element regexes. Added a regression test pinning 14/16 stays 14/16 through the round trip.

   While verifying #2 I found the same "editor locks itself out on its own output" class in two more places your list mentioned in passing - fixed both, since they're reachable through the actual UI:
   - Custom text filter excluded any line containing the bare words "pris"/"$"/"img" (not just the actual markup), so "Pris pr. stk" or "Kun $99" in a free-text custom line was silently dropped. Narrowed to the real tokens.
   - width/height/font_size/margin_top/margin_left are free-text fields; "3,5" produced invalid CSS in the first place and couldn't be read back either. Normalized to a dot at generation time (plus a comma-tolerant read fallback for anything already stored with one).

3. Agreed, leaving as-is - this is the accepted trade-off, not a bug. Worth noting the four shipped templates are the only ones this still gates now that #1/#2 are fixed (the "everything's blocked" symptom is gone for anything the visual editor could have produced or that predates it via the 4.0 migration).

4. Added the escape hatch to the warning text itself (tekst 5059): "...for at få adgang til den visuelle editor igen skal den rå HTML tømmes og gemmes."

5. Fixed properly rather than just inverting the condition - extracted the whole guard→generate→save flow into `saveVisualLabelEdit()` (sys_div_func.php), which both diverse.php and the tests now call, so there's one code path instead of two that could drift. Refusal now sets `$saveLabelRefused` and `labels()` shows an explicit "not saved, reload and retry" message (tekst 5060) instead of silently re-rendering.

6. This is what the `saveVisualLabelEdit()` extraction fixes - `attemptGuardedVisualSave()` in the test now calls it directly instead of reimplementing the branch inline, and I added explicit assertions on its return value (refused/applied) to both existing tests using it, plus built the POST array to match a real unchecked-checkbox submit (absent key, not `false`) rather than the convenience shape the helper used internally. That surfaced a real `if_isset()` eager-dereference warning in the extracted function itself (same bug class as a couple other tickets recently) - fixed as part of the extraction.

**Should fix** - all done: `findtekst()` for the warning strings (5059, 5060 in tekster.csv) instead of the hardcoded Danish literal; the `$saveLabel` block is now properly tab-indented instead of mixed with the old nested if/else; `preg_replace` now collapses whitespace to a single space instead of stripping it entirely (was silently treating templates that differ in internal spacing as identical); docblock has `@param`/`@return`, and `# endfunc` was on the wrong function (belonged to `labels()`, not `labelTemplateEditableVisually()`) - fixed both, plus added the marker to the two other new functions in this area for consistency with the rest of the file.

Full characterization suite (16 tests now, was 9) green locally, including 4 new regression tests for the font-size/custom-text/decimal-comma fixes. Ready for another look whenever you have time - I know freeze is tomorrow 18:00.

ZaynSaul:
Quick note: master moved again right after my last push (3 more commits landed upstream, including #465). Merged and pushed (076d4b3e) - back to mergeable, no conflicts.

One side effect: #465 added its own tekster.csv rows in the 5059-5077 range, colliding with the two IDs (5059/5060) I'd used for the messages mentioned above. Renumbered mine to 5078/5079 and updated both `findtekst()` calls in sys_div_func.php to match - same text, just different IDs now. Full suite still green.

SirRolin (approved):

PR #510, opened 2026-08-26
Updated order location from finance to debitor

Loe4g:
finans/ordre.php has gone stale and is no longer being maintained, while debitor/ordre.php has become the up-to-date, canonical implementation for order handling. This PR updates the faktura column's link in kontospec.php to point to ../debitor/ordre.php instead of the legacy ordre.php in finans/, aligning it with the POS branch (pos_ordre.php) which already lived in debitor/.

Since the target file has moved to a different directory, the returside parameter (used by ordre.php, pos_ordre.php, and kassekladde.php to link back) was also updated from kontospec.php to ../finans/kontospec.php so the return navigation still resolves correctly from debitor/.

Why: consolidates order-viewing logic onto a single, actively maintained file (debitor/ordre.php) rather than continuing to support two divergent implementations.

No changes to grid columns, filters, or query logic — this is purely a link-target correction. tied to PR #498 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated transaction and invoice links to navigate to the correct pages.
  * Improved return navigation to the account overview after completing invoice or cash journal actions.


ZaynSaul (approved):
Approving despite the missing "how I verified this" line — I did the verification independently:

- Confirmed finans/ordre.php is a genuine stale 2010 fork (its own file header still literally reads "debitor/ordre.php ... lap 3.0.1 ... 2010-05-20"), while debitor/ordre.php is the actively maintained one.
- Confirmed debitor/ordre.php correctly reads $_GET['id'] and $_GET['returside'].
- Checked debitor/ordre.php's load path for any write triggered by a bare ?id=&returside= visit — every db_modify() there is gated behind additional GET/POST keys this link never sets (vis_lev_addr, kontakt/email, sort+fokus+funktion, POST linjetekster), so this is read-only in practice.
- This mirrors the already-verified POS branch from PR #498 (../debitor/pos_ordre.php), just extended to the non-POS case.
- php -l clean, no conflicts with master (7 commits behind, none touching kontospec.php).

Two non-blocking nits left as FYI in the earlier comment (stray trailing whitespace, unrelated kassekladde.php returside tweak) — not holding up merge for those.

SirRolin (approved):
approved by ZaynSaul

LuiHedlund02 (approved):

PR #513, opened 2026-08-27
F-002: kassekladde.php plain-load warning flood (init pass, no behavior change)

LuiHedlund02:
## Background

`finans/kassekladde.php` is one of the largest warning sources in every release-gate run (200+ hits per run on ~15 source lines) — finding F-002 in the August test sweeps. Each warning is a candidate fatal under a future PHP upgrade, and the volume drowns real errors in the PHP log. This is a warning-silencing pass only: **no behavioral change**.

## Changes

- **`if_isset($_GET['...'])` → `if_isset($_GET, null, '...')`** for `visipop`, `udskriv`, `tjek`, `returside`. The array-first form evaluates the missing key before the helper runs, which is exactly the warning the helper exists to prevent (see the usage doc in `includes/std_func.php:175`). Return values are identical.
- **Initialise `$kksort` / `$kkdir` / `$returside` to `null`** at the top. They are read (header sort links at ~2543, `build_kassekladde_query` at ~2082, back-link at ~1496) before the KASKL defaults load at ~2604. `$kkdir` is deliberately initialised to `null`, not `'asc'` — the sort-toggle links compare `== 'asc'`, so `'asc'` would flip their direction; `null` keeps behaviour byte-identical.
- **Initialise `$antal = 0`** in the POST intake. `$antal` is never assigned anywhere in the file, so the "insert balance line after last line" block at ~1220 (`$x == $antal - 1`) has never run. `0` keeps it not running, warning-free. **The inert auto-balance block itself is a separate defect** (SD-617-adjacent) and should get its own ticket — deliberately not activated here because that would change posting behaviour.
- **`??` guards** on `$dokument[$y]`, `$id[$y]`, and `$bilag[$x-1]` reads in the row-render loops, matching the `?? ''` style the surrounding 2025 code already uses. Missing keys already behaved as null; now they do so without warning. (Note: `$y` in the new-row clip block at ~3205 is stale from the previous loop — that is pre-existing and untouched.)

## Verification

- `php -l` clean.
- Every changed expression evaluates to the same value as before for the undefined case (see the `$kkdir` note above for the one trap).

Part of the F-002/F-003/F-011 warning sweep (testy findings ledger); sibling PRs fix `debitorkort.php` and `ordreliste.php`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cash journal stability when optional values or form inputs are missing.
  * Prevented warnings and unexpected behavior when processing document attachments and voucher numbers.
  * Improved handling of empty or undefined data during cash journal operations.

### DIFF
--- a/finans/kassekladde.php
+++ b/finans/kassekladde.php
@@ -461,10 +461,13 @@ if (!isset($tidspkt))
 	$tidspkt = 0;
 if (!isset($row['tidspkt']))
 	$row['tidspkt'] = null;
+$kksort    = null;
+$kkdir     = null;
+$returside = null;
 
-$visipop = if_isset($_GET['visipop']);
-$udskriv = if_isset($_GET['udskriv']);
-if ($tjek = if_isset($_GET['tjek'])) {
+$visipop = if_isset($_GET, null, 'visipop');
+$udskriv = if_isset($_GET, null, 'udskriv');
+if ($tjek = if_isset($_GET, null, 'tjek')) {
 	$tidspkt = microtime();
 	list($a, $b) = explode(" ", $tidspkt);
 	$qtxt = "select bogfort,tidspkt,hvem from kladdeliste where (bogfort = '-' or bogfort = 'S') and id = $tjek";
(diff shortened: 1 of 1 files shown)


ZaynSaul (approved):
Thoroughly verified — independently checked all three non-trivial claims in the description:

1. if_isset($_GET['key']) vs if_isset($_GET, null, 'key') — confirmed via if_isset()'s source: the old form evaluates $_GET['key'] as a PHP argument before the function runs (exactly where the warning fires); the new form passes array+key so the function's own array_key_exists() check avoids it.
2. $kkdir initialized to null, not 'asc' — confirmed by tracing its read at line 2543 ($_next_dir = ($kkdir=='asc') ? 'desc':'asc'), which runs before the real KASKL default load at ~2606. Initializing to 'asc' instead would have silently flipped the sort-toggle's initial direction — correctly avoided.
3. $antal = 0 — confirmed $antal has exactly one live reference in the file ($x == $antal - 1), never assigned elsewhere; with $x always >= 1 this stays permanently false either way, so the dead auto-balance block stays dead.

php -l clean, no conflicts with master, no CodeRabbit findings. Approved.

(Noting for the record: this touches finans/kassekladde.php, which is on your own stated escalation list — but since you're the author here, I'm treating your own review-queue placement as the call already made.)

PR #515, opened 2026-08-27
F-011: ordreliste.php per-row sprog_id warning + show_shop_link init

LuiHedlund02:
## Background

The order/invoice list warns once per displayed row — the single biggest warning source in a release-gate run (finding F-011, found by the AI UI tester 2026-08-10). Two causes:

1. The `sum` column render closure imports `global $genberegn` but not `$sprog_id`, so the `findtekst('1442|Ikke udlignet', $sprog_id)` tooltip call warns on **every faktura row** and always falls back to the default language instead of the user's configured one. The sibling closure at ~762 already imports `$sprog_id` correctly.
2. `$show_shop_link` is only assigned inside the shop-API config branch (~1986), so every load without a shop integration warns at the footer check (~2645).

## Changes

- `global $genberegn, $sprog_id;` in the `sum` render closure.
- `$show_shop_link = false;` before the shop-integration block.

Behavioral note: the "Ikke udlignet" tooltip now renders in the user's configured language on non-Danish accounts — that is the bug being fixed; everything else is byte-identical.

## Verification

- `php -l` clean.

Part of the F-002/F-003/F-011 warning sweep (testy findings ledger).

### DIFF
--- a/debitor/ordreliste.php
+++ b/debitor/ordreliste.php
@@ -1089,7 +1089,7 @@ $custom_columns = array(
         "decimalPrecision" => 2,
         "searchable" => true,
         "render" => function ($value, $row, $column) use ($valg, &$ialt, &$ialt_m_moms, &$ialt_kostpris) {
-            global $genberegn;
+            global $genberegn, $sprog_id;
 
             if ($genberegn) {
                 $kostpris = genberegn($row['id']);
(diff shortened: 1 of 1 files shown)


ZaynSaul (approved):
Verified both claims directly:

1. Line 1124 (findtekst('1442|Ikke udlignet', $sprog_id)) sits inside the exact "sum" render closure being patched, which previously only imported global $genberegn — not $sprog_id. Confirms the claimed per-row warning + silent fallback to default language.
2. $show_shop_link is only ever assigned at line 1986, inside the shop-API config branch; read unconditionally at line 2645. Without the shop integration configured, it was undefined there. The new $show_shop_link = false; before the block fixes this cleanly.

php -l clean, no conflicts with master.

Note for whoever merges: this file is also touched by #512 (different region — lines ~2650+ there vs ~1089/~1982 here), so no expected conflict between the two, but flagging since both are in-flight on the same file.

Approved.

PR #526, opened 2026-08-28
Fix: SD-660 $tpm/$tmp typo, missing prev/next bounds guards, and unset $opVa…

ZaynSaul:
## Summary
Opening an existing order/invoice in `debitor/ordre.php` emitted three PHP warnings. On
investigation, these turned out to be two warning-only defects and one real, silently-broken
behavior: the post-print redirect target has not been correctly set since a prior commit
accidentally deleted the line that set it.

### Issue 1 — $tpm/$tmp typo (line 3723)
`if($tpm != $value_type) $tmp = $value_type;` — `$tpm` never existed anywhere in the
codebase; it was always a typo for `$tmp`. Confirmed via `git blame`/history: commit history
at this line reads "SD-368: Sets tmp to value_type if set for navigation," so `$tpm` was
never intentional. The condition was therefore always true, and the guard it was meant to
implement never guarded.

**Fix:** `$tpm` → `$tmp`.

**Behavior impact:** none. Traced where `$tmp`/`$value_type` actually flow — they only ever
fed the already-commented-out OLV lookup two lines below. This is a warning-only fix;
confirmed prev/next links are identical before/after on all three test orders (first,
middle, last).

### Issue 2 — missing prev/next bounds guards (lines 3734/3736)
`if ($ordrlst[$i - 1])` / `$ordrlst[$i + 1]` had no bounds checks — reading key `-1` on the
first element, and reading past the end on the last. The intended NULL-at-the-ends behavior
worked today only by accident.

**Fix:** added `isset()` guards around both array accesses.

**Behavior impact:** none. Live-reproduced the exact warning (`Undefined array key -1`) on
the first order in the list pre-fix; confirmed the NULL-at-the-boundary behavior is
byte-identical before/after (same rendered arrow links on first/middle/last).

### Issue 3 — unset $opValue silently broke the post-print redirect (line ~4140)
`$opValue` was read in the plukliste condition
(`($hurtigfakt == 'on' && $opValue == 'faktura') || ...`) without being set on all paths.
This was more than a warning: `git log -S` traced it back to commit `92dd1295`
("Live Pickaxe... #356"), which merged two `if`/`elseif` plukliste blocks into one combined
condition and accidentally deleted the line that originally set
`$opValue = $_GET['valg']` and wrote it to `temp/$db/area$bruger_id.txt`. That file is read
by `includes/udskriv.php:423` to pick the post-print redirect target (faktura vs.
tilbud/ordrer).

**This has been silently broken since commit `92dd1295` merged** — the post-print redirect
target has not been correctly written since.

**Fix:** restored the assignment at its original location, using
`if_isset($_GET, NULL, 'valg')` instead of raw `$_GET['valg']` to avoid reintroducing a
warning.

**Behavior impact:** real. Live-confirmed: before the fix, loading an order never wrote
`area-1.txt`; after the fix, `?valg=faktura` correctly writes `faktura` into it, restoring
the intended post-print redirect behavior.

## Files Changed
- debitor/ordre.php

## Out of Scope
- The order-line save path and anything touching posting

## How to Test
1. Open the first, middle, and last order of the faktura list
   (`debitor/ordre.php?id=...`) — verify zero PHP warnings from `ordre.php` in the error log
   for all three.
2. Verify prev/next navigation arrows render and behave identically at both boundaries (no
   broken links, no change in target on first/middle/last).
3. Open an order with `hurtigfakt` on and off, and with `valg` set to `faktura` vs.
   `tilbud`/`ordrer` — verify the plukliste block appears/disappears under exactly the same
   conditions as before this fix.
4. Print/finalize an order with `?valg=faktura` and confirm `temp/$db/area$bruger_id.txt`
   is written with the correct value, and the post-print redirect lands on the correct
   target page.
5. Repeat step 4 for `tilbud`/`ordrer` values and confirm the redirect target differs
   correctly per value.
6. Confirm all request input touched in this fix (`$_GET['valg']`) is safely read via
   `if_isset()`, not raw.

## Acceptance Criteria
|---|---|---|
| 1 | Opening first, last, and a middle order of a list | Zero PHP warnings from ordre.php; prev/next navigation behaves identically at the boundaries |
| 2 | Plukliste block under hurtigfakt on/off × faktura/tilbud | Appears/disappears under exactly the same conditions as before |
| 3 | Request input in touched code | Int-cast or db_escape_string()ed; no raw $_GET/$_POST in SQL |
| 4 | Dead/commented-out code | None; one ticket per PR |

## Verification
Verified live against `chartest`, first/middle/last order of a 42-order faktura list,
before vs. after (via `git stash`):
- All three named warnings reproduced pre-fix, gone post-fix.
- No other new warnings introduced (remaining log noise is pre-existing, out-of-scope —
  `$konto_id`, `$viewport_width`, etc.).
- Plukliste block presence unchanged (didn't render on these three orders either way — they
  are case-linked (`sag_id` set), which the code already excludes).
- Prev/next arrow targets identical before/after.
- Test DB temp files (`ordrlst-1.txt`, `area-1.txt`) cleaned up after verification.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored proper handling when processing order pages.
  - Fixed navigation between orders to prevent incorrect behavior.
  - Added safeguards to avoid errors when previous or next orders are unavailable.
  - Corrected an issue affecting the active user context when a selection is provided.
- **Documentation**
  - Updated the associated change documentation with additional details about the fix.


ZaynSaul:
@coderabbitai fixed in c21a3b46 — the file write is now guarded on `$opValue !== NULL`, so it only overwrites `area$bruger_id.txt` when the request actually carries `valg`; otherwise the previously-remembered value is left alone (which is exactly what the invoice form's post-without-valg needs).

Live-verified against a real tenant DB: seeded the area file via `?valg=faktura`, then loaded the same order with no `valg` param — confirmed the file still reads `faktura` on the fixed code, and confirmed it gets blanked on the pre-fix code via `git stash` (i.e. reproduced the exact bug you flagged, then confirmed it's gone).

LuiHedlund02 (approved):
Reviewed the current rebased diff: the conflict was confined to history comments. The existing live reproduction covers boundary navigation and preserving the remembered print destination when valg is absent; PHP lint and diff checks pass on the rebased head.

PR #533, opened 2026-09-01
Fix: Kontokort report shows identical Saldo on every line (MB-34)

LuiHedlund02:
## Problem
Finans → Rapporter → Kontokort: the Saldo column shows the same value (the primosaldo) on every line for an account instead of a running balance. The CSV export has the same defect.

## Cause
When pagination was added, the running-balance line (`$kontosum += debet - kredit`) was moved into the `rows_to_skip` branch in `finans/rapport_includes/kontokort.php`. It therefore only ran for rows skipped by pagination — never for the rows actually printed, so every printed line repeated the balance as of the start of the page.

## Fix
Hoist the accumulation out of the skip branch so it runs for every transaction row, printed or skipped. This restores the original semantics (each line's Saldo includes its own posting): the debet/kredit cells print from the row values directly, and the saldo cell — including the currency-account branch dividing by `transkurs` — reads `$kontosum` after the row is applied. Cross-page consistency is unchanged: the next page still re-derives its opening balance by summing the skipped rows.

## Test
- Kontokort for an account with many postings (e.g. 25100 Debitorer): Saldo now changes line by line; last visible line on page 1 equals the opening context of page 2.
- Page 2+ of a paginated account: first line's Saldo continues from where page 1 ended.
- Currency account (kontovaluta set): Saldo still converted via transkurs, tooltip shows DKK value.
- CSV export: Saldo column shows the same running values as the screen.

https://claude.ai/code/session_01RUrafXy5qSgAsBnzuefqcA

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected running-balance calculations in account statements.
  * Balances now remain accurate across paginated and displayed transactions.

### DIFF
--- a/finans/rapport_includes/kontokort.php
+++ b/finans/rapport_includes/kontokort.php
@@ -31,7 +31,10 @@
 // 20210301 PHR error in csv.
 // 20250130 migrate utf8_en-/decode() to mb_convert_encoding
 // 20260309 LOE Fixed execessive error logging relating to undefined array keys.
-// 20260430 LOE Updated the top menu and made the report header sticky when scrolling. 
+// 20260430 LOE Updated the top menu and made the report header sticky when scrolling.
+// 20260901 CL/LAH Fixed Saldo column showing the same value on every line:
+//                  running balance was only accumulated for rows skipped by
+//                  pagination, never for the printed rows.
 
 function kontokort($regnaar, $maaned_fra, $maaned_til, $aar_fra, $aar_til,
                    $dato_fra, $dato_til, $konto_fra, $konto_til, $rapportart,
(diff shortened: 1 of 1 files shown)


ZaynSaul (approved):
Gate-check passes (Problem/Cause/Fix/Test description is thorough, CodeRabbit approved with no findings, checks green, single commit).

Full review: traced $kontosum through the whole lifecycle in finans/rapport_includes/kontokort.php.
- Reset correctly per account from primo[$x] (+ prior-year folding) before the per-transaction loop - no cross-account leakage.
- The fix makes accumulation unconditional (runs before the skip check instead of only inside it), so it now covers both paginated-skip rows and printed rows, with no double-counting since the skip branch no longer re-adds it.
- The currency-account branch (division by transkurs) and the CSV fwrite both read the same corrected $kontosum, so both get fixed together, matching the PR's own test note.
- Cross-page continuity holds: page 2+ still re-derives its opening balance by summing every prior page's now-correctly-accumulated skipped rows.
- Only one print loop in the file - no parallel/legacy code path was missed.

php -l clean. No existing characterization tests for this file to run. Approved.

SirRolin (approved):

PR #538, opened 2026-09-02
Kassekladde: ignore a replayed save - double-click / back + resend (L4 adversarial-forms DEVY-2, adversarial-navigation DEVY-1)

LuiHedlund02:
Repeated saves of the same journal form could insert the same lines twice. A new journal was also allocated before the old replay check, so double-clicking the first save created two journals.

Completed saves are now checked immediately after tenant authentication, before journal creation or staging. A per-form identity and the exact submitted payload scope the replay cache to the tenant and user; duplicates redirect to the saved journal. Failed first saves retain their allocated journal so a retry or correction does not allocate another one. Save validation failures are not recorded as successes. Separately rendered forms and changed payloads remain valid saves. Other actions do not use this cache.

Merged current master (`e9442a65`). The session cache retains the most recent 128 completed saves and 128 form-to-journal mappings; this is session-scoped replay protection.

## Validation

- `php tests/integration/journalSaveReplay.php`: 27 checks, with warnings enabled.
- Actual HTTP requests against PHP 8.3/PostgreSQL in an isolated disposable tenant: simultaneous first saves produced one journal and one row (200/303); Back/resend preserved one row; ordinary edits persisted; replaying an older completed request preserved a later edit; an independent identical new form created a separate journal; failed retries and subsequent correction reused one journal.
- The original branch head reproduced two journals/two rows under the same simultaneous-first-save test.
- PHP lint and whitespace checks passed.

## Reviewer scenarios

In `finans/kassekladde.php`, repeat double-click and Back/resend on both new and existing journals. Save an invalid amount twice, correct it, and verify the error and typed lines remain available until a successful save. Then save, edit, save, and resend the older request: the later edit must survive. Verify lookup, copy, simulate, and posting still perform their ordinary actions.


SirRolin (dismissed):

SirRolin (changes requested):
When I create a new kassekladde and quickly double click "gem" it saves 2 identical kassekladder, but the 3rd empty one doesn't show up.

LuiHedlund02:
@SirRolin please re-review the current head after the September 14 follow-up. Focus on the save-token/concurrent-submit behavior from your review and confirm the follow-up resolves it; no additional code was pushed in this handoff.

PR #540, opened 2026-09-02
Self-heal adresser.stripe_fravalg where the debtor card writes it (L4 master-data-customer DEVY-1)

LuiHedlund02:
## What

L4 gate 2026-09-02, **master-data-customer DEVY-1** and **concurrent-editing DEVY-1** (no SD ticket yet): every debtor save/create failed with `column "stripe_fravalg" of relation "adresser" does not exist` and a blank page. Whole customer master-data was dead on those tenants.

The migration exists in `includes/betweenUpdates.php`, but that file only runs when a regnskab is opened through `admin/aaben_regnskab.php` (or the POS). A tenant reached any other way never gets the column, while `debitorkort.php` / `debkort_save.php` write it unconditionally. Same lesson as F-014 (`global_id`) and SD-646 (`moms_periode_luk`).

## How

- `includes/stdFunc/ensureColumn.php` (new): `ensure_column($table, $column, $definition)` — one `information_schema` lookup per request, `ADD COLUMN IF NOT EXISTS` (two concurrent logins can both pass the check, same reason as the `ansatte.mobile` guard).
- Called from `debitor/debitorkort.php` and `debitor/debkort_save.php` before the writers.
- `betweenUpdates.php` now uses the same helper, so the migration path stays canonical and there is one implementation.

## Review focus

Deploy-checklist item as well: a scripted pre-cutover schema-parity assertion (`stripe_fravalg`, `moms_periode_luk`, `global_id`) would catch this class before customers do.

## Verification

- `php -l` clean on all five files (via the `saldi-web` image).
- Not yet exercised in the test lab. Acceptance: fresh clone that has *not* been through `aaben_regnskab.php` — open a debtor card and save; row persists and the column exists afterwards.

https://claude.ai/code/session_01XGPZkSoY2Li3KuckJu6N7C

### DIFF
--- a/debitor/debitorkort.php
+++ b/debitor/debitorkort.php
@@ -116,6 +116,11 @@ $css = "../css/standard.css";
 include("../includes/connect.php");
 include("../includes/online.php");
 include("../includes/std_func.php");
+// 20260902 CL/LH  L4 finding master-data-customer DEVY-1 / concurrent-editing DEVY-1: the debtor
+// INSERT/UPDATE below reference adresser.stripe_fravalg unconditionally, but the column is only
+// added by includes/betweenUpdates.php (aaben_regnskab.php). Self-heal here so a tenant that
+// skipped that path does not lose every debtor save to "column does not exist".
+ensure_column('adresser', 'stripe_fravalg', 'varchar(2)');
 include("../includes/topline_settings.php");
 include("../includes/grid.php");
 include_once("../includes/emballage_schema.php");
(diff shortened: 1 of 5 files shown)


SirRolin (changes requested):
I generally like the introduction of ensure_column for the use in betweenUpdates or opdat files, but using them like that directly in the files means that we might get conflicts in the future or simply outdated "requirements".
Instead we should check if there's a difference between the version number of the database and the code and relog the user if different.

LuiHedlund02:
Closing per the September triage decision. Schema repair belongs in the canonical update path; this request-time self-heal does not address Nicolai’s objection.

PR #542, opened 2026-09-02
finans/bogfor.php: never create a zero-amount open post (L4 dunning-run DEVY-5)

LuiHedlund02:
## What

L4 gate 2026-09-02, **dunning-run DEVY-5** (no SD ticket yet): finalising a zero-total reminder inserted a second, unsettled `openpost` row for the customer with `faktnr=1`, amount `0.000`, next to the real overdue invoice 1 — a duplicate open item under the original invoice number.

## How

`finans/bogfor.php` `openpost()`: return early when `|amount| < 0.005`. A zero row can never settle anything either — the `±0.005` match window only ever matches other zero rows — so nothing legitimate is lost.

## Review focus (money path — senior review, @sirrolin)

Both callers (`bogfor()` debit and credit legs, lines ~827/832) pass the line amount; a genuinely zero line now creates no open post on either side. If there is a flow that relies on a 0,00 open post existing, this is the place to say so.

## Verification

- `php -l` clean (via the `saldi-web` image).
- Not yet exercised in the test lab. Acceptance: repeat dunning-run without a fee — `openpost` for konto 1535 must still contain only the original invoice row; a reminder *with* a fee must create exactly one open post for the fee.

https://claude.ai/code/session_01XGPZkSoY2Li3KuckJu6N7C

## Summary by CodeRabbit

* **Bug Fixes**
  * Zero-value postings below 0.005 are no longer matched with or added to open items.


SirRolin (approved):
We do use 0 amount kassekladder, but according to claude, the later reads filter 0 posts away anyways so this should just reduce data, with no cosequences.

PR #553, opened 2026-09-04
MB-35: unify Indkøbsforslag reorder formula across both calculation b…

ZaynSaul:
## Summary
`lager/varer.php`'s "Indkøbsforslag" (reorder suggestion) feature computes, for each item,
how much to reorder. Its `udskriv()` function had two separate code branches computing this
differently:

- One branch (decides *whether* an item needs reordering) correctly subtracted existing
  purchase proposals and open purchase orders (`i_forslag`/`bestilt`) already covering part
  of the gap.
- The other branch (renders the actual editable quantity shown to the user) used a simpler
  formula that ignored `i_forslag`/`bestilt` entirely.

So an item already covered by a pending purchase order could still be suggested for the full
gap up to max — effectively proposing a duplicate order on top of one already in flight.

## ⚠️ Reproduction caveat — read before treating this as confirmed root cause of the reported "17"
The ticket reports a specific observed value (genbestillingsforslag 17 where the expected
value is 9, for stock 9 / min 5 / max 10 / open sales order 8). **This exact "17" figure was
not reproducible with clean data on any code path in this investigation.** What *was*
confirmed and fixed is a real, deterministic inconsistency between the two branches
described above — this is a genuine bug and this PR fixes it. But it should not be assumed
this PR explains the specific "17" the QC reviewer observed; that may have a different or
additional cause (the ticket itself raises two competing hypotheses: a `varer.beholdning`
desync, or order-line double-counting in `find_beholdning()`). Neither of those was confirmed
or ruled out here — see "Not Investigated" below.

**Recommendation:** verify with whoever filed the original ticket whether the underlying
`ssl3` dev-mappen scenario can still be reproduced post-fix, or whether the desync/
double-counting hypotheses need separate follow-up.

## Note on file location
The ticket assignment message pointed at `lager/minmaxstock.php` for this formula — that
file does not contain this logic at all. The actual code is in `lager/varer.php`, matching
the ticket description's own line references (line 910 / 1001-1003). Flagging this in case
it indicates stale documentation/tribal knowledge about where this logic lives.

- Extracted two small shared functions — `genbestil_nettobeholdning()` and
  `beregn_genbestil()` — into `includes/std_func.php`, next to `find_beholdning()` (whose
  values they consume).
- Routed both branches in `lager/varer.php` through these shared functions instead of each
  repeating the formula inline, so they can no longer drift apart.
- Added a PHPUnit test suite with the ticket's worked example (stock 9, min 5, max 10, open
  sales order 8 → expected 9), plus cases that specifically exercise the branch
  inconsistency this fix resolves, and a zero-clamp case (reorder suggestion should never go
  negative).

## Files Changed
- lager/varer.php
- includes/std_func.php
- tests/characterization/includes/genbestil/GenbestilCharacterizationTest.test.php (new)

## Not Investigated (flagged, not ruled out)
Per the ticket's own two hypotheses for the specific "17" figure:
1. `varer.beholdning` desync from `sum(lagerstatus.beholdning)` — not confirmed or ruled out.
2. Order-line double-counting in `find_beholdning()` (`includes/std_func.php:1598`) — not
   confirmed or ruled out.

Neither was reproducible with clean test data. If the original "17" is still observed
post-fix in the real `ssl3` scenario, one of these is the more likely remaining cause and
should be investigated as a follow-up.

## How to Test
1. Run the worked example from the ticket: item with stock 9, min lager 5, max lager 10, one
   open (undelivered) sales order for 8 units. Verify the Indkøbsforslag now shows 9, not 17.
2. Create an item with an existing open purchase order or purchase proposal already covering
   part of the reorder gap. Verify the visible suggested quantity now correctly subtracts
   that pending order/proposal, matching the "needs reordering" branch's calculation.
3. Confirm both branches (the reorder-needed decision and the visible suggested quantity)
   now agree for the same item in all test cases.
4. Verify the zero-clamp case: an item that doesn't actually need reordering (net position
   already at or above max) shows a suggested quantity of 0, not a negative number.
5. Run the new PHPUnit suite and confirm all cases pass, including the ticket's worked
   example.
6. Attempt to reproduce the original ticket's exact "17" scenario against the `ssl3`
   dev-mappen demo account, if still available, and note whether it's resolved or whether a
   different root cause is still present.

## Acceptance Criteria
|---|---|---|
| 1 | Worked example: stock 9, min 5, max 10, open sales order 8 | Genbestillingsforslag = 9 |
| 2 | Item with existing open purchase order/proposal covering part of the gap | Visible suggested quantity correctly subtracts it — no duplicate-order suggestion |
| 3 | Both calculation branches for the same item | Consistent with each other |
| 4 | Item not needing reorder | Suggested quantity is 0, never negative |

## Verification
- The worked example now correctly computes 9 (not 17) both in the unit test and live,
  against a test scenario in the local `chartest` tenant.
- **Not verified:** the literal "17" from the ticket was not reproducible with clean data on
  any code path — see caveat above.

## Summary by CodeRabbit

- **New Features**
  - Reorder suggestions now calculate net stock using current inventory, pending purchase proposals, and existing purchase orders.
  - Suggested reorder quantities are calculated against the configured maximum stock level.

- **Bug Fixes**
  - Reorder suggestions now consistently account for pending purchasing activity across all supported workflows.
  - Negative reorder quantities are prevented, ensuring suggestions never recommend less than zero items.
  - Reorder calculations now provide consistent results regardless of the workflow used to generate them.

### DIFF
--- a/includes/std_func.php
+++ b/includes/std_func.php
@@ -75,6 +75,9 @@
 //                 deleted account's number is never reused; kontonr above 8 digits (EAN-like
 //                 outliers) are ignored when finding the highest, and if the 8-digit range is
 //                 capped by an outlier the first number free in both series is used (SST-753)
+// 20260905 SZ Added genbestil_nettobeholdning()/beregn_genbestil() next to find_beholdning() -
+//             lager/varer.php's reorder-suggestion formula had two branches that disagreed on
+//             whether existing purchase proposals/orders were subtracted (MB-35)
 
 include(__DIR__ . '/stdFunc/dkDecimal.php');
 include(__DIR__ . '/stdFunc/nrCast.php');
@@ -1675,6 +1678,24 @@ if (!function_exists('find_beholdning')) {
 		return $beholdning;
 	}
 } #endfunc find_beholdning()
+
+// MB-35 - lager/varer.php's reorder-suggestion code had two branches computing this differently:
+// one subtracted existing purchase proposals/orders ($i_forslag/$bestilt, from find_beholdning()
+// above) from both the trigger and the suggested quantity, the other used neither anywhere, so an
+// item already covered by a pending purchase order still got flagged and suggested for the full
+// gap up to max - doubling the order once the pending one also arrives. Both branches now share
+// these two functions so they can't drift apart again.
+if (!function_exists('genbestil_nettobeholdning')) {
+	function genbestil_nettobeholdning($beholdning, $i_ordre, $i_forslag, $bestilt) {
+		return $beholdning - $i_ordre + $i_forslag + $bestilt;
+	}
+}
+if (!function_exists('beregn_genbestil')) {
+	function beregn_genbestil($max_lager, $beholdning, $i_ordre, $i_forslag, $bestilt) {
+		$genbestil = $max_lager - genbestil_nettobeholdning($beholdning, $i_ordre, $i_forslag, $bestilt);
+		return $genbestil < 0 ? 0 : $genbestil;
+	}
+}
 
 if (!function_exists('hent_shop_ordrer')) {
 	function hent_shop_ordrer($shop_ordre_id, $from_date)
--- a/lager/varer.php
+++ b/lager/varer.php
@@ -57,7 +57,12 @@
 // 2021.04.01 LOE - Translated these texts to English 20210401
 // 2023.04.14 LOE - Minor modifications
 // 2023.06.03 PHR - php8
-// 2023.09.05	PHR - cookie for saldiProductListStart & saldiProductListLines 
+// 2023.09.05	PHR - cookie for saldiProductListStart & saldiProductListLines
+// 20260905 SZ MB-35: udskriv()'s two genbestil branches disagreed on whether existing purchase
+//             proposals/orders (i_forslag/bestilt) were subtracted from the reorder suggestion -
+//             one branch used neither, so an item already covered by a pending purchase order got
+//             suggested for the full gap to max again. Extracted genbestil_nettobeholdning()/
+//             beregn_genbestil() and routed both branches through them.
 
 @session_start();
 $s_id=session_id();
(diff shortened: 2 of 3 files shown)


LuiHedlund02:
Lead reviewer: @SirRolin (already requested)
Reason: changes the suggested reorder quantity for every item with open purchase proposals or orders.
Focus: confirm with the MB-35 reporter that netting `i_forslag`/`bestilt` is the wanted rule; the reported value 17 was never reproduced, so this may not close the complaint.
Verification: 5-case PHPUnit provider on the pure helpers; no integration test through `udskriv()`.
Next action: @ZaynSaul resolve conflicts; review can start on the diff now.

LuiHedlund02:
@ZaynSaul please resolve conflicts and confirm with the reporter that netting i_forslag/bestilt is the intended reorder calculation; the reported 17 was not reproduced. Then request @SirRolin to review the formula with a concrete stock/order example and expected result.

ZaynSaul:
@SirRolin conflicts are resolved and merged past master (test suite still green: 10/10 on `GenbestilCharacterizationTest`). Since you reported MB-35, a couple of things before this merges:

1. **Does netting `i_forslag`/`bestilt` match what you expected?** The fix makes both calculation branches subtract existing purchase proposals and open purchase orders already covering part of the gap, so an item with a pending order in flight no longer gets suggested for the full amount again. Want to confirm this is the intended rule before it changes the suggested quantity for every item that has open proposals/orders.

2. **The reported "17" wasn't reproducible.** For the worked example (stock 9, min 5, max 10, open sales order 8), this fix correctly gets 9 - live-verified locally, and covered in the test suite. But I couldn't reproduce the specific "17" figure from the ticket on any code path with clean data, so I can't confirm this fix is what actually explains what you originally saw. If the `ssl3` dev-mappen scenario that produced the 17 is still around, could you re-check it against this branch? If it's still off, the ticket's other two hypotheses (a `varer.beholdning` desync, or order-line double-counting in `find_beholdning()`) haven't been ruled out and would need separate follow-up.

Once you've had a chance to weigh in on both, could you also review the formula itself with a concrete stock/order example and expected result? Lui asked for your review here specifically since this changes the suggested reorder quantity for every item with open purchase proposals or orders.

ZaynSaul:
@SirRolin important follow-up before this merges - I think I found what actually produces "17", and it's not in this PR's diff.

**`lager/varer.php`'s own Indkøbsforslag button appears unreachable for real customers.** I checked this live against a restored real customer database: `lager/varer.php` unconditionally redirects to `lister/vareliste.php` whenever the session's menu preference is `'S'` (the modern UI) - which is what that customer's session is set to. The `udskriv()` function this PR fixes never actually renders; the request redirects away before reaching it.

**The live Indkøbsforslag experience is a separate file with its own separate implementation.** `lister/vareliste.php`'s "Indkøb" button goes to `lager/lister/indkøb.php`, which computes the same suggestion with its own independent SQL query - not through `genbestil_nettobeholdning()`/`beregn_genbestil()`. I checked the whole codebase: those two functions (what this PR fixes) are referenced nowhere except `lager/varer.php` and its own test file.

**`indkøb.php`'s formula already nets in_sales_order/in_buy_proposal/in_buy_order correctly** - it doesn't have this PR's branch-inconsistency bug at all. But it adds something neither `varer.php` nor the ticket's own worked example accounts for: it rounds the suggestion up to a full case/pack size (`volume_lager`, "Pakningsmængde" in the code): `CEIL((max - net) / volume_lager) * volume_lager`. For the ticket's own numbers (net=1, max=10, gap=9): if that item's case size is 17, `CEIL(9/17) * 17 = 17` - exactly the reported figure. If that's what happened, "17" isn't a bug - it's the correct case-rounded quantity, and "expected 9" just didn't account for packaging rounding.

I couldn't confirm this against the exact original data point (today's live stock on that tenant has moved past the scenario), but the mechanism is real and lines up exactly.

**Question for you:** which page were you looking at when you saw 17 - `lager/varer.php`'s own list, or the "Indkøb" page (`lister/vareliste.php` → `indkøb.php`)? If it was the latter, this PR's fix - correct as it is for the code it touches - doesn't reach the page or the code path where 17 was actually observed, and MB-35 would need a second PR against `indkøb.php` (and a decision on whether 17 was ever actually wrong).

PR #556, opened 2026-09-05
MB-38: assign next free ordrenr to new order created via Opret ny ord…

ZaynSaul:
## Summary
Splitting a supplier order and choosing "Opret ny ordre" (create new order) as the split
target cloned the source order's number (`ordrenr`) onto the new order instead of assigning
it a fresh one — so two (or more) supplier orders ended up with the identical order number.
This broke anything that looks orders up by number: `kreditor/ordre.php`'s own
ordrenr-based lookup, supplier-invoice matching, and the purchasing overview.

**This is very likely the root cause of previously archived production issue SST-237**
(MEDSHOP: "kreditorordre oprettet 2 gange med samme kreditorordrenummer" — a 0,00 duplicate,
reported happening "from time to time" in production, archived without a confirmed root
cause). Worth reopening/cross-referencing SST-237 once this is verified.

Related: MB-1 / SST-652 (#525) fixed blank pages in this same file — this is the numbering
sibling of that issue.

## Root Cause
In `kreditor/orderIncludes/moveOrderLines.php`, the "create new order" path copies the
source order row into a temp table, rewrites only the `id` column, then inserts it back into
`ordrer`:

```sql
CREATE TEMPORARY TABLE temp_table AS SELECT * FROM ordrer WHERE id='$id'
UPDATE temp_table SET id='$newId' WHERE id='$id'
INSERT INTO ordrer SELECT * FROM temp_table
```

Every other field, including `ordrenr`, gets copied over unchanged — nothing later in the
file corrects it.

Before building the copy, compute the next free kreditor order number and set it on the temp
row alongside the rewritten `id`. Used the same scoping the codebase already relies on
elsewhere for this (`kreditor/ublimport.php`):

```sql
select max(ordrenr) from ordrer where art = 'KO' OR art = 'KK'
```
+1, then set on the temp row before the insert.

Confirmed via the database that kreditor (`KO`) and debitor (`DO`) orders use separate,
overlapping numbering — a plain global `max(ordrenr)` across all order types would have
picked the wrong (potentially still-colliding) number.

## Files Changed
- kreditor/orderIncludes/moveOrderLines.php

## How to Test
1. Creditors → Order → open an order with 2+ lines.
2. Click "Split" → in the target dropdown choose "Opret ny ordre", set a quantity to move,
   click Flyt.
3. Verify the newly created order has a **different** ordrenr from the source order — not a
   clone.
4. Open the split dialog again on the same source order and repeat the split — verify the
   dropdown does not show a duplicate order number, and each new split-created order gets
   its own distinct, incrementing number.
5. Check Creditors → Order list — verify no two orders share the same ordrenr.
6. Verify number-based lookups still work correctly for both the original and the newly
   created order: `kreditor/ordre.php`'s ordrenr lookup, supplier-invoice matching, and the
   purchasing overview.
7. Confirm both `KO` (order) and `KK` (credit note) order types are correctly included in
   the next-number calculation, and that a `DO` (debitor) order number is never
   accidentally reused or collided with.
8. Reproduce the original bug against pre-fix code (e.g. via `git stash`) to confirm it
   still occurs there, then confirm it's resolved post-fix — proving this is a genuine
   fix, not a coincidental non-reproduction.

## Acceptance Criteria
|---|---|---|
| 1 | Split a supplier order using "Opret ny ordre" as target | New order gets the next free ordrenr, not the source's number |
| 2 | Repeat split multiple times on the same or different source orders | No two orders ever share the same ordrenr |
| 3 | Number-based lookups (kreditor/ordre.php, invoice matching, purchasing overview) | Work correctly for all resulting orders |
| 4 | KO/KK numbering scope | Correctly separated from DO (debitor) numbering — no cross-type collision |

## Verification
Verified live end-to-end: built a test order and performed the actual split through the
real running app.
- Pre-fix: the new order came out with the same order number as the source, reproducing the
  exact reported bug.
- Post-fix: the new order got the next free number instead, with no duplicates left
  anywhere.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved safeguards when splitting orders and creating new orders.
  * New split-off orders now receive unique order numbers more reliably, even when multiple orders are created concurrently.
  * Order creation stops safely if protection against duplicate order numbers cannot be secured.
  * Prevented potentially duplicate order numbers after a failed protection check.
  * Transactions are rolled back when the required protection cannot be established.

### DIFF
--- a/kreditor/orderIncludes/moveOrderLines.php
+++ b/kreditor/orderIncludes/moveOrderLines.php
@@ -42,6 +42,9 @@
 //             splits could still compute the same "next" number. Serialize new-order creation with
 //             an advisory lock (pg_advisory_xact_lock, auto-released on commit; GET_LOCK/RELEASE_LOCK
 //             on MySQL, released explicitly after commit since MySQL locks aren't transaction-scoped).
+// 20260905 SZ MB-38/CodeRabbit: GET_LOCK() returns 0 on timeout / NULL on error, which the earlier
+//             fix ignored - check for a return of 1 and abort the split with an alert instead of
+//             risking a duplicate ordrenr if the lock wasn't actually acquired.
 
 print "<!-- BEGIN orderIncludes/moveOrderLines.php -->";
 #print "moveOrderLines.php<br>";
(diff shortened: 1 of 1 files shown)


SirRolin (approved):
Tested on test_12 by creating a new orderline, adding 2 items, splitting it to a new order instead of existing.

LuiHedlund02:
Pre-deploy review. The split path now gets a fresh number, good. Two things before merge and one follow-up:

1. **Check the lock result.** `db_select("SELECT pg_advisory_xact_lock(hashtext('kreditor_ordre_split_ordrenr'))", ...)` (`moveOrderLines.php:97`) ignores the return value. With a `statement_timeout` set, a contended lock fails; `db_select()` alerts and continues (and on current master its `ROLLBACK` from #503 ends the transaction, see #560), so the header insert and line moves run outside the transaction without the lock. Abort before allocating if the select fails.
2. **MySQL `GET_LOCK` name is server-wide** (`'kreditor_ordre_split_ordrenr'`, line ~87): a slow split in tenant A blocks tenant B. Prefix the name with the tenant DB. (Postgres fleet today, but the code path exists.) Also the alert text has `l&aring;st` / `Pr&oslash;v` entities inside a JS string; use the literal characters.
3. **Follow-up, not for this PR:** CodeRabbit's heavy lift is real. `kreditor/ublimport.php:337`, `lager/varer.php` and `kreditor/scanAttachments/frontpage/frontTable.php` allocate KO/KK numbers without this lock, so an import paused between `MAX(ordrenr)+1` and its insert can still collide with a split. Worth an SD ticket and a cross-reference to SST-237.

PR #557, opened 2026-09-07
Fix credit-note return date validation

momocoder14:
A customer credit note must accept a return date earlier than its creation date. The same applies to an all-negative customer order that becomes a credit note during posting. Ordinary sales and purchase orders must still reject delivery dates before the order date.

The shared date check in `includes/ordrefunc.php` validates the effective order type in all four paths: debtor approval, finance approval, delivery, and posting. Empty dates remain invalid. Positive, mixed-sign, empty, and zero-quantity orders do not receive the credit-note exception.

Merged current master (`e9442a65`), resolving the history-header conflict while retaining the current employee-field changes and the credit-note approval fix.

## Validation

- `php tests/integration/creditNoteDateValidation.php`: 44 checks execute the actual four guard expressions and credit-note classifier with a SQL double, with warnings enabled.
- `OrderDateValidationCharacterizationTest.test.php`: 4 tests / 4 assertions passed.
- PHP lint and whitespace checks passed.
- These checks validate classification and guards; they do not execute the complete accounting posting transaction.

## Reviewer scenarios

In `debitor/ordre.php`, `finans/ordre.php`, and `includes/ordrefunc.php`, approve/deliver/post a DK credit note and an all-negative DO order with a return date before the order date. Repeat with an ordinary positive order, mixed-sign lines, missing dates, and a purchase order: those must retain their date restrictions. Check the employee fields after the merge.

The branch also requires `levdate` in the quick-invoice delivery path where the previous quick-invoice bypass allowed it to be absent. This pre-existing change in the PR remains a specific review point; the new regression matrix exercises normal delivery, not the full quick-invoice workflow.

## Summary by CodeRabbit

- **Bug Fixes**
  - Credit notes can now use return dates earlier than the original order date when applicable.
  - Orders with missing delivery dates are now validated consistently.
  - Improved VAT validation prevents incorrect posting totals and identifies the affected account and order.
  - Approval, delivery, and posting checks now consistently recognize orders that will become credit notes.
- **User Experience**
  - Packing-slip actions now return users to the relevant order.
  - The “Completed by” field is positioned beside “Our reference” for easier viewing.

### DIFF
--- a/debitor/ordre.php
+++ b/debitor/ordre.php
@@ -107,6 +107,7 @@
 // 20260818 Sawaneh Credit notes: only cap the quantity when the line points the wrong way or more
 //                  is credited than invoiced, so it can be reduced. Handles invoice lines that are
 //                  themselves negative. Shows the max in the alert. Removed debug_kreditnota logging.
+// 20260831 CDX/MJ JOB-106 Allow credit-note return dates before the credit-note order date
 // 20260907 CDX/LH Share the invoice payment gate with the assistant's saved-state reader.
 // 20260908 CL/Sawaneh SST-763: PBS button on posted PBS invoices opens debitor/pbs_gensend.php
 //                     (attempt history + resend after a Nets rejection).
@@ -1649,7 +1650,7 @@ if (($status < 3 || strstr($b_submit, "Kopi") || strstr($b_submit, "Kred")) && $
 				print "<BODY onLoad=\"javascript:alert('$alert')\">\n";
 				$levdate = date("Y-m-d");
 			} else $levdate = $ordredate;;
-		} elseif ($levdate < $ordredate) {
+		} elseif (delivery_date_before_order_date($art, $levdate, $ordredate)) {
 			$alert1 = findtekst('1679|Leveringsdato er før ordredato', $sprog_id);
 			print "<BODY onLoad=\"javascript:alert('$alert1')\">\n";
 			$status = 0;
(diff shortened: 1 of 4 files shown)


LuiHedlund02:
Pre-deploy review: the helper and the four call sites look right, and the characterization test covers the helper. One thing before merge, which is CodeRabbit's open Major: all-negative `DO` orders are converted to `DK` **after** these validations run, so both paths can still reject a valid return date for that case. Pass the effective type (`$dan_kn ? 'DK' : $art`) to `delivery_date_before_order_date()` at `includes/ordrefunc.php:1445`, and classify before `levering()` validates at `:267`.

Optional: the tests only exercise the helper; an integration case that approves and delivers a `DK` with return date < order date (plus a `DO` control that is still rejected) would catch a caller regression.

SirRolin (changes requested):
marking as requested changes.

LuiHedlund02:
Re-review requested: @momocoder14 merged master and pushed the credit-note date validation on 2026-09-14, after the changes-requested of 2026-09-08. @SirRolin

LuiHedlund02:
@momocoder14 your September 14 follow-up is recorded, but the refreshed queue still reports master conflicts. Please resolve them and re-run credit-note date checks before @SirRolin’s re-review. Preserve his existing blocking review and request; do not treat the later push alone as acceptance.

momocoder14:
@LuiHedlund02 The new code rabbit finding is real but pre-existing and out of scope, with the ticket link.

PR #562, opened 2026-09-07
Fix: varespor grouping key, adjustment sign and missing cell (#495 follow-up)

LuiHedlund02:
## What
Three defects in the grouping added by #495 (`lager/varespor.php`):
1. Rows with `linje_id = 0` (every stock correction from `lager/lagerregulering.php` and `lager/lagerflyt.php`) were grouped by `modtime` + warehouse. On MySQL `modtime` has second granularity, so two independent movements of the same item saved in the same second collapsed into one line with a doubled quantity and price.
2. `abs()` on `antal` (lines ~157 and ~250) discarded the stored sign, so reversals were shown and totalled with the wrong sign in `$adjustmentTotal`.
3. The else branch that printed the firmanavn cell was deleted, so a Tilgang row whose order was deleted emitted one `<td>` fewer than the header and shifted the price column.

## How
Fallback key is `row:<id>` (only rows sharing a real `linje_id` merge); signed quantities kept (batch_salg still negated); empty right-aligned cell restored. Lines 167-173 re-indented one level (they were one tab too deep inside the loop).

## Manual test
1. Item with two stock corrections saved in one submit: two lines, not one doubled line.
2. Item with a negative correction: the Lagerreguleringer total goes down, not up.
3. Item bought on an order that was later deleted: columns still align.

Cross-checked by GPT-6 Astra (correct; pre-existing warnings for a deleted order remain).

https://claude.ai/code/session_01J4xFKPi2G5K4dYXtY8ZrhV

### DIFF
--- a/lager/varespor.php
+++ b/lager/varespor.php
@@ -25,6 +25,8 @@
 // 20140626 Tilføjet lagerregulering og ændret variabelnavn for dækningsbidrag.
 // 20150902	Linjer med 0 i antal undertrykkes og linjer uden ordre_id vises som Lagerreguleret
 // 20260819 CDX/PHR Saml fragmenterede batchlinjer, vis lager og saml lagerreguleringer nederst.
+// 20260907 CL/LH Saml kun rækker med samme linje_id (ikke modtime), bevar fortegn på
+//                lagerreguleringer og genopret tom firmanavn-celle ved slettet ordre (#495).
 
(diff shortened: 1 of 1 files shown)


SirRolin (approved):

SirRolin (approved):
Tested all Manual Tests in Debitor.

PR #551, opened 2026-09-04
REST API: broaden the PHPUnit suite, seed a random test password, fix 3 endpoint bugs

SirRolin:
Tests (tests/restapi, all on RestApiEnv against the throwaway tenant):
- AuthRefreshEndpointTest: refresh token -> working access token; access, garbage, closed/unknown-account, unknown-user and claim-less tokens refused; X-Tenant-ID fallback.
- BearerAuthEnforcementTest: the shared BaseEndpoint gate (no header, Basic, refresh-typed, foreign-secret, expired, unknown account, malformed JSON).
- CustomersEndpointTest, ProductsEndpointTest, AccountsEndpointTest, OrderLinesEndpointTest: create -> read-back, validation, search, update, delete, 404/405, creditor vs debtor separation, posted-order refusal.
- ReferenceDataEndpointsTest: currencies, vat, vat-codes, accountingYear, debitor/groups, products/groups, dashboard/stats need a token, answer the JSON envelope, refuse writes. Suite: 29 -> 108 tests; skips cleanly without the stack.

RestApiEnv: authHeaders/loginData/refreshToken/regnskabId/signToken helpers, login cache reset per bootstrap, and teardownTenant() called from tearDownAfterClass() in every class so the seeded login never outlives a class (SALDI_REST_KEEP_TENANT=1 keeps it for debugging).

Secrets: the seeded API user's password is no longer a literal. The tracked tests/TestCredentials.php generates it at random per process (SALDI_TEST_PASSWORD_<KEY> overrides it for logging in by hand). The old literal is in git history and is treated as public. doc/ai/convention_no_hardcoded_secrets.md tightened accordingly: seeded test users are no longer exempt, must use the random per-run password, and must be torn down. tests/README.md added; tests/restapi/README.md updated.

Bugs the new tests exposed, fixed minimally:
- vat-codes/index.php and dashboard/stats.php redeclared BaseEndpoint's protected $db as private - a PHP fatal, so every request returned an HTML error page.
- CustomerModel::save() synced an empty contact-email list back to adresser.email on insert, wiping the e-mail just supplied, so API-created customers had no e-mail and the duplicate check never matched.

Not fixed, for a ticket: products/groups and inventory/warehouses list empty under JWT because global $regnaar is only set on the legacy API-key path; an access token stays valid after its account is closed.

## Have you checked the following? 
- [x] I have updated relevant documentation at https://github.com/DANOSOFT/saldi/wiki
[...]

PR #564, opened 2026-09-07
opdat: reopen the dead 4.3.0 gate and stamp tenants to 5.0.0; REST notifications/register fatal; vareliste redirect (deploy-gate blocker)

LuiHedlund02:
## What
From the 2026-09-03 pre-deploy review ("Saldi 5.0.0 Deploy Gate"), two commits:

**1. `includes/opdat_4.2.php`, `includes/opdat_4.3.php`**
The 4.3.0 schema step sat behind `if ($subNo < '1')`, which `tjek4opdat()` can never satisfy (it always passes `$subNo = 2`), so `adresser.fax -> mobile`, `kontakt_emails`, `delivery_addresses` etc. were skipped while the code already depends on them. Gate on `< '3'`; every statement inside is existence-guarded, so re-running on a partially upgraded tenant is safe. `opdat_4.3.php`: drop the early return, reconnect to the tenant after the 4.2.x steps left `$connection` on the master DB, and let `opdat_to()` stamp tenants from 4.3.0 to the installed 5.0.0 so login stops re-entering the update chain.

**2. `restapi/endpoints/v1/notifications/register.php`, `lager/vareliste.php`**
register.php had the same PHP 8 fatal as dashboard/stats and vat-codes (#551): a `private` redeclaration of BaseEndpoint's protected `$db`/`$userId`. Also answers GET with 405. vareliste.php (2004-era, short open tags) now redirects legacy bookmarks to `varer.php`, forwarding only the supported `sort`/`vis_lev` options.

The three REST fixes that #551 carries (stats, vat-codes, CustomerModel) were dropped from this branch.

## Manual test
1. Tenant at 4.2.x: open regnskab, confirm `adresser.mobile`, `kontakt_emails`, `delivery_addresses` exist afterwards and the grupper VE row says 5.0.0.
2. Tenant already at 4.3.0: login stamps to 5.0.0, no errors, second login does not re-run the chain.
3. `POST /restapi/v1/notifications/register` with a valid token: 200 JSON, no HTML error page; GET returns 405.
4. `lager/vareliste.php?sort=varenr` redirects to `varer.php?sort=varenr`.

https://claude.ai/code/session_01J4xFKPi2G5K4dYXtY8ZrhV


SirRolin (changes requested):
if I click dg or manually change sort[varelst] to something that's not on the list, I get 2 fatal errors.
The following sortings should also be supported: varegruppe, stregkode, leverandør, lager_total, momspris, kortpris, dg

LuiHedlund02:
@SirRolin addressed the sorting report and rebased. Master already contains the 4.2 gate correction, which is retained. `apply_sort_sqlOverride()` now rejects unknown/malformed fields before SQL; the item list sorts DG by the selected `dg` alias, fixing the DISTINCT/ORDER BY mismatch. Supported-column regression: 14 ascending/descending cases plus 5 invalid inputs pass; two pre-existing grid signature deprecations remain visible. The column is named `kostpris` (the review said `kortpris`). Please re-test varegruppe, stregkode, leverandør, lager_total, momspris, kostpris and DG, and a fabricated `sort[varelst]` field. Also re-check the notification endpoint and a 4.3.0 tenant upgrade in a disposable environment. #551 and this PR have disjoint changed-file sets; there is no file overlap.

PR #575, opened 2026-09-08
SD-558: fix optional performed-by selection and Danish label

LuiHedlund02:
The customer order's “Performed by” field can disappear when no active employees exist, and its name is still hardcoded in English. An early customer update can also clear the saved performer before POST input is read; names containing quotes are interpolated directly into SQL and HTML.

Complete the SD-558 follow-up: always render an optional dropdown with blank selected by default, preserve historical names, allow explicit clearing, and preserve the stored value when the field is absent. Remove the obsolete performer-based takeover fallback, escape performer values at SQL/HTML boundaries (including copy and split paths), and use findtekst for “Udført af” / “Performed by” / “Utført av” on the order and order list.

Jira: https://saldi-team.atlassian.net/browse/SD-558
Related original feature: SD-523; label request: MB-3.

Validation: PHP 8.3 syntax checks passed for all changed application PHP files. PHPUnit 11: 5 tests, 19 assertions passed with E_ALL, covering rendered options, special-character round trips, absent-versus-blank save input, and the early customer update. git diff --check passed. Full browser/database order workflows have not been exercised.

Manual regression checks before merge:
- debitor/ordre.php and orderIncludes/performedBy.php: create/save/reopen with blank; select another technician, save/reopen, clear, save/reopen. Repeat with no active employees and a historical/inactive performer. Selection must never trigger an order-takeover message.
- Order persistence and read-only/offer views: use a name containing apostrophes, quotes, &, and Danish letters; confirm exact display and persistence. A form action omitting hvem must preserve it.
- Copy, credit, and split order actions: smoke-test an order with an apostrophe in the performer; confirm each action completes and retains its existing copy/reset semantics.
- debitor/ordreliste.php and importfiler/tekster.csv: verify Danish and English labels and displayed performer after returning to the list, including after clearing.

## Summary by CodeRabbit

- **New Features**
  - Added an optional “Performed by” selection when creating and editing orders.
  - The selector includes a blank option, active employees, the current user, and previously saved names without duplicate entries.
  - Previously saved performer values remain available and can be cleared.

- **Bug Fixes**
  - Improved handling and display of performer names, including special characters.
  - Updated the order list to use the translated “Performed by” label consistently.


LuiHedlund02:
@ZaynSaul rebased, allocated translation ID 5231, and folded the remaining changes into master’s existing `renderEmployeeFields.php` instead of adding another renderer. The searchable list column uses the same translation. Missing performer input preserves the stored value; explicit blank clears it. The existing concurrent-edit warning is retained. Both performer/employee suites pass: 11 tests, 49 assertions; PHP lint passes. Decision: this is the focused performer fix; #447 must separate its competing locking design from this UI work and #589. Manual checks: blank/default/historical employee, apostrophe name, missing-field save, copy/credit/split and posted display.

ZaynSaul (approved):
Traced the three claimed fixes directly in the diff, all confirmed:

1. **Early customer-update clobbering `hvem`** - the customer-info UPDATE statement no longer includes `hvem = '...'` at all; the new test (`testCustomerUpdateDoesNotOverwritePerformerBeforePostIsRead`) extracts that exact SQL-building block via string search and asserts it contains no `hvem` write, testing the real production code rather than a reimplementation.
2. **SQL/HTML escaping** - `db_escape_string((string)$r['hvem'])` added to the copy-order INSERT (previously interpolated raw), and `htmlspecialchars((string)$hvem, ENT_QUOTES, 'UTF-8')` added at three separate HTML output sites (two hidden inputs, one read-only display). The round-trip test uses a name containing `O'Brien "<script>" & Æøå` and asserts zero `<script>` elements parse out - a real proof, not just a description.
3. **Missing-vs-blank distinction** - `$hvem = isset($_POST['hvem']) && is_string($_POST['hvem']) ? $_POST['hvem'] : null` correctly distinguishes "field absent" (preserve stored value) from "field present but empty" (explicit clear), and the later `if ($hvem === null)` load-from-DB branch matches. Tested against null/''/'0'/a quoted name.

No text-ID collision: checked current master for 5231, none exists (549's 5230 and this PR's 5231 are coordinated, non-colliding).

The PR body is upfront that full browser/DB workflows haven't been exercised yet - only PHPUnit-level coverage (which is unusually rigorous here, extracting real code blocks via eval rather than testing a parallel reimplementation). Worth doing the listed manual regression checks before deploy, but the logic itself is well-verified. Approving.

PR #578, opened 2026-09-08
Fix Danish login redisplay and missing input-length helper

PeterRude:

When an account cannot be found during login, Danish characters in the account name and username can be replaced with entity fragments such as `oslash` when the login form is redisplayed. Pass the original values to `login()`, which already handles HTML escaping and sanitization, instead of first encoding them with `htmlentities()`.

Follow-up to #573. Login also counts characters directly with `mb_strlen()` instead of calling the new `is_input_too_long()` helper. This preserves the existing 60/80-character limits and prevents the reported fatal error on installations where the loaded `std_func.php` does not provide that helper.

Validation:
- `php -l index/login.php` and `git diff --check` passed.
- Exercised the unknown-account call and existing sanitizer with `Øresund æøåÆØÅ`: account and username were preserved in UTF-8 and ISO-8859-1.
- The reporter confirmed the initial Danish-character fix works; subsequently reported the missing-helper fatal error on another installation.
- Tested the sanitizer with no `is_input_too_long()` function defined: Danish characters and the 60/61-character account and 80/81-character username boundaries passed in UTF-8 and ISO-8859-1.

Manual regression scenarios:
- `index/login.php`: submit a nonexistent account containing æøåÆØÅ and confirm the account and username remain intact on return to the login page.
- `index/login.php` / `index/index.php`: use an incorrect password with Danish account and user names; confirm the fields remain intact and a corrected login succeeds.
- Confirm successful login with Danish characters in the account, username and password.
- On the installation reporting the missing-helper error, verify login reaches authentication without a fatal error and still rejects account names above 60 characters and usernames above 80 characters.

- [ ] I have updated relevant documentation at https://github.com/DANOSOFT/saldi/wiki (not required for this bug fix).
- [x] I have checked that I am not linking to any external resources such as external style or JavaScript files that could compromise stability.
- [ ] I have read and understood the linked developer guidelines (repository agent instructions and project memory were followed).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved login input handling for Danish characters.
  * Corrected input length validation for multibyte characters.
  * Preserved entered login values when displaying unknown-account errors.

### DIFF
--- a/index/login.php
+++ b/index/login.php
@@ -60,6 +60,7 @@
 //                  back to the page charset so a non-UTF8 database still matches. Length check now uses the
 //                  shared is_input_too_long() from std_func.php.
 // 20260908 CDX/PHR Preserve Danish characters when redisplaying an unknown account.
+// 20260908 CDX/PHR Count login input characters directly to support older std_func.php installations.
 
 ob_start(); //Starter output buffering 
 @session_start();
@@ -202,7 +203,7 @@ function sanitize_input($input, $allowed_length = 80) {
 		return false;
 	}
 
-	if (is_input_too_long($input, $allowed_length)) {
+	if (mb_strlen($input, 'UTF-8') > $allowed_length) {
 		return false;
 	}
 
--- a/index/login.php
+++ b/index/login.php
@@ -59,6 +59,7 @@
 //                  posted from an ISO-8859-1 page is filtered instead of rejected, and the result is converted
 //                  back to the page charset so a non-UTF8 database still matches. Length check now uses the
 //                  shared is_input_too_long() from std_func.php.
+// 20260908 CDX/PHR Preserve Danish characters when redisplaying an unknown account.
 
 ob_start(); //Starter output buffering 
 @session_start();
@@ -342,7 +343,7 @@ if (isset($_POST['regnskab'])) {
 		exit();
 		}
 		if ($regnskab) $fejltxt="Regnskab $regnskab findes ikke";
-		login(htmlentities($regnskab,ENT_COMPAT,$charset),htmlentities($brugernavn,ENT_COMPAT,$charset),$fejltxt);
+		login($regnskab,$brugernavn,$fejltxt);
  	}
 } else {
 	


LuiHedlund02:
Lead reviewer: @ZaynSaul
Reason: login redisplay fix plus inlining `is_input_too_long()`.
Focus: the helper is on master at `includes/std_func.php:2949` with six other callers from #573; an install without it still fatals at those six sites. Which install reported the fatal, and is inlining the fix or finishing that deploy? The escaping half checks out: `login()` already does `htmlspecialchars` + `sanitize_input`.
Verification: php -l, boundary tests at 60/61 and 80/81, UTF-8 and ISO-8859-1 redisplay; no automated test.
Next action: review now (mergeable).

LuiHedlund02:
@ZaynSaul retaining your review request for login redisplay and character-length handling. Focus: Danish account names redisplay without double encoding, and `mb_strlen` matches the shared helper. @PeterRude please identify the installation that lacked `is_input_too_long()`; master contains it and other callers still depend on it, so partial deployment needs explaining. Test unknown Danish account names and maximum-length username/account inputs.

ZaynSaul (approved):
Reviewed the login redisplay fix. Traced both root causes against the actual code:
1. The old htmlentities() call before login() double-encoded input - login() already does its own htmlspecialchars()+sanitize_input() pass, and sanitize_input's whitelist strips '&' and ';', so an entity like &oslash; lost its delimiters and left literal "oslash" text on screen. Removing the outer htmlentities() call is the correct fix and matches the other 3 call sites in the file that already pass raw values.
2. The missing-helper fatal is fixed by inlining is_input_too_long()'s exact logic (verified byte-for-byte equivalent against includes/std_func.php) rather than depending on a function that older installs may not have.

No password exposure (login()'s signature has no password param, untouched here), no XSS regression (same single-htmlspecialchars-then-strip pattern as before). Approved - login flow changes warrant a real click-through before merge: submit an unknown account with æøåÆØÅ and confirm clean redisplay, on both charset variants.

PR #583, opened 2026-09-09
SST-759: Explain rejected recipient characters and offer explicit ASC…

sawaneh1:
…II correction

SST-759: Explain rejected recipient characters and offer explicit ASCII correction

Background: MEDSHOP (saldi_390) could not mail invoice 240666 because the order's
e-mail held U+2212 MINUS SIGN instead of an ASCII hyphen. The old rejection only
said "Invalid email format in ...", and the address looked correct on screen.

Changes
- New includes/formFuncIncludes/emailLookalike.php: analyses a rejected address
  character by character and reports code point, Unicode name, 1-based position
  and the ASCII replacement for known lookalikes (dashes, invisible whitespace,
  fullwidth forms, Greek/Cyrillic homoglyphs). Unknown non-ASCII characters are
  still identified by code point but no suggestion is made.
- sendMail.php and oldSendMail.php: rejection goes through the shared feedback.
  The alert is JSON-encoded and page output is HTML-escaped, which also closes
  the unescaped alert("... $emails[$x]") in the old code.
- Explicit acceptance only: in the debitor print popup, single-mail flow, when
  the rejected address is literally on ordrer.email, the page stays open with the
  offending character marked, the suggestion, and an accept button. Batch flows
  (reminders, statements) and kreditor keep the old alert-and-return behaviour.
- debitor/formularprint.php: POST email_fix_from handler checks debitor rights,
  re-derives the suggestion server-side from the rejected address (client cannot
  supply an arbitrary address), verifies the order still holds that exact
  address, updates only that recipient, then redirects to the e-mail print URL.
- Texts 5210-5221 in tekster.csv. No migration needed.
- tests/test_email_lookalike.php: 21 CLI assertions, no DB required.

Manual tests on ssl12 (saldi_34, order 3342)
- dintest@agf−as.dk (U+2212): alert names MINUS SIGN, U+2212, position 12,
  suggests dintest@agf-as.dk; popup stays open with character highlighted.
- Accept button: "changed to dintest@agf-as.dk", order e-mail field now shows
  ASCII hyphen. ssl12 has no outbound mail, so delivery itself was not observed.
- php tests/test_email_lookalike.php passes locally and on the server.

Incident status
- Manual UPDATE on saldi_390 order 162742 applied 2026-09-09 (1 row affected).
- Customer confirmation of a successful resend still pending.

Not covered / out of scope
- Silent bulk cleanup, IDN/international addresses, unsolicited customer mail.

## Summary by CodeRabbit

* **New Features**
  * Invalid email addresses now identify problematic characters and provide suggested ASCII replacements.
  * Users can review and apply suggested corrections when printing emails.
  * Corrections are validated for permissions and stale-address conflicts before being applied.

* **Bug Fixes**
  * Improved handling of email addresses containing lookalike, invisible, or non-ASCII characters.
  * Added support for recognizing legacy character encodings during validation.

* **Tests**
  * Added coverage for character detection, address suggestions, encoding conversion, and invalid addresses.

### DIFF
--- a/debitor/formularprint.php
+++ b/debitor/formularprint.php
@@ -31,6 +31,7 @@
 // 20260102 LOE Added department support for background files
 // 20260309 PHR Fixed error in $returside after printing
 // 20260309 PHR Fixed another error in $returside after printing
+// 20260909 Sawaneh SST-759: POST email_fix_from applies the recipient suggestion from send_mails() after explicit acceptance
 
 
 session_start();
@@ -88,6 +89,26 @@ function find_background_file($background, $file_type, $department) {
     }
 
     return null;
+}
+// 20260909 Sawaneh SST-759: the user accepted the suggested recipient correction shown by
+// send_mails(); the suggestion is re-derived server-side from the rejected address.
+if (isset($_POST['email_fix_from']) && isset($_POST['id'])) {
+    $id = (int)$_POST['id'];
+    $formular = (int)if_isset($_POST['formular']);
+    if ($rettigheder && substr($rettigheder, 5, 1) < '1') {
+        print tekstboks(findtekst('5221|Du har ikke rettigheder til at ændre ordrens e-mailadresse', $sprog_id));
+        exit;
+    }
+    include_once(__DIR__ . "/../includes/formFuncIncludes/emailLookalike.php");
+    $newEmail = emailLookalikeApply($id, $_POST['email_fix_from']);
+    if ($newEmail === '') {
+        print "<script type='text/javascript'>alert(" . json_encode(findtekst('5216|Ordrens e-mailadresse er ændret siden - kontrollér og send igen', $sprog_id), JSON_INVALID_UTF8_SUBSTITUTE | JSON_HEX_TAG) . ");</script>";
+        print "<meta http-equiv=\"refresh\" content=\"0;URL=ordre.php?id=$id\">";
+        exit;
+    }
+    print findtekst('5220|E-mailadressen på ordren er rettet til', $sprog_id) . " " . htmlspecialchars($newEmail, ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8') . "<br>";
+    print "<meta http-equiv=\"refresh\" content=\"1;URL=formularprint.php?id=$id&formular=$formular&udskriv_til=email\">";
+    exit;
 }
 //check Post for returside
 $returside = if_isset($_POST['returside']);
--- a/importfiler/tekster.csv
+++ b/importfiler/tekster.csv
@@ -5096,3 +5096,15 @@
 5149	Brugernavnet må højst være 80 tegn	The username may be at most 80 characters	Brukernavnet kan høyst være 80 tegn
 5150	Regnskabsnavnet må højst være 60 tegn	The account name may be at most 60 characters	Regnskapsnavnet kan høyst være 60 tegn
 
+5210	Ugyldig e-mailadresse	Invalid e-mail address	Ugyldig e-postadresse
+5211	Ugyldigt tegn	Invalid character	Ugyldig tegn
+5212	position	position	posisjon
+5213	Foreslået rettelse	Suggested correction	Foreslått rettelse
+5214	Brug den foreslåede adresse og send igen	Use the suggested address and send again	Bruk den foreslåtte adressen og send på nytt
+5215	Ret adressen manuelt på ordren eller debitorkortet	Correct the address manually on the order or the debtor card	Rett adressen manuelt på ordren eller debitorkortet
+5216	Ordrens e-mailadresse er ændret siden - kontrollér og send igen	The order's e-mail address has changed since - check and send again	Ordrens e-postadresse er endret siden - kontroller og send på nytt
+5217	Ukendt tegn	Unknown character	Ukjent tegn
+5218	erstattes med	is replaced by	erstattes med
+5219	fjernes	is removed	fjernes
+5220	E-mailadressen på ordren er rettet til	The e-mail address on the order was changed to	E-postadressen på ordren er rettet til
+5221	Du har ikke rettigheder til at ændre ordrens e-mailadresse	You do not have permission to change the order's e-mail address	Du har ikke rettigheter til å endre ordrens e-postadresse
--- a/includes/formFuncIncludes/emailLookalike.php
+++ b/includes/formFuncIncludes/emailLookalike.php
@@ -25,6 +25,8 @@
 // 20260909 Sawaneh SST-759: Explain why a recipient address is rejected (character,
 //                  code point, position), suggest the ASCII equivalent for known
 //                  lookalikes and let the user apply it explicitly on the order.
+// 20260910 Sawaneh SST-759: Apply only when the stored recipient is unchanged since it was
+//                  shown (conditional update + re-read), per CodeRabbit review on PR #583.
 
 if (!function_exists('emailLookalikeTable')) {
 	/**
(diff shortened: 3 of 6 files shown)


LuiHedlund02:
Lead reviewer: @ZaynSaul
Reason: new POST that updates `ordrer.email` from a shared mail path.
Focus: the rights gate is fail-open (`if ($rettigheder && substr(...) < '1')` skips the check when `$rettigheder` is empty); `emailRecipientRejected()` calls `exit` inside a shared mail function, confirm the batch flows (reminders, statements) cannot strand a partly written job.
Verification: 21 CLI assertions, manual on ssl12 order 3342; outbound delivery not observed.
Next action: review now (mergeable). Text ids 5210-5221 stay with this PR; #585 renumbers.

LuiHedlund02:
@ZaynSaul retaining your lead for recipient lookalike correction. Review the fail-open shape of the rights gate when `$rettigheder` is empty and the effect of `emailRecipientRejected()` exiting from the shared mail path. Author reports 21 fixture assertions; verify explicit acceptance, stale address rejection, no unauthorized writes and cancellation without sending. This PR keeps text IDs 5210–5221; #585 must renumber before merge.

ZaynSaul (dismissed):
Reviewed SST-759. The one real CodeRabbit finding (a concurrent-update race in emailLookalikeApply that could overwrite a newer edit) is fixed with a proper compare-and-swap pattern plus read-after-write staleness check - verified in the current diff.

Checked independently for header injection, SQL injection, and XSS: all clean. CR/LF are in the lookalike table and stripped, plus filter_var(FILTER_VALIDATE_EMAIL) already rejects raw CR/LF before this code runs. Queries are int-cast/escaped. All dynamic output goes through htmlspecialchars with ENT_QUOTES|ENT_SUBSTITUTE, and this also fixes a pre-existing unescaped alert() as claimed.

Danish letters (æøå) are deliberately left as "unknown, no suggestion" rather than guessed at - documented as an explicit scope choice (IDN/international addresses out of scope), not an oversight.

Approved - the actual Accept-and-resend flow still needs a live end-to-end check per the PR's own notes (ssl12 has no outbound mail to verify delivery against).

LuiHedlund02:

ZaynSaul (approved):
Ran the test suite on the branch (21/21 pass, matches the PR's claim). Read the full 391-line lookalike table — genuinely broad coverage (Unicode minus/dashes, NBSP/ZWSP variants, soft hyphen, fullwidth forms, ~25 Greek/Cyrillic homoglyphs), not just the one reported character, and unrecognized non-ASCII is reported rather than guessed.

Confirmed no silent auto-mutation: the fix is only reachable via an explicit POST from a real form, re-derives the suggestion server-side (client can't supply an arbitrary replacement), re-checks the address is still on the order, and uses an optimistic-concurrency WHERE-clause guard plus a re-read to confirm the write — closes the CodeRabbit concurrency finding for real, not just per the commit message. HTML/JS escaping is properly centralized through a local htmlspecialchars()/json_encode() closure, replacing the old raw string-interpolated alert().

Checked both of your own open questions directly: the fail-open rights gate when $rettigheder is empty mirrors the exact same guard shape already used in includes/online.php's core per-page access gate, so it's existing convention rather than a new hole — worth a one-line FYI, not a blocker. And the exit()-inside-shared-mail-function concern doesn't strand batch sends: traced $mailantal through both the invoice and reminder mail loops and confirmed it's set once to the full batch count before the loop runs, so the interactive single-recipient branch can only fire when there's truly nothing left in the loop to strand. LGTM.

ZaynSaul:
Follow-up to my own reviews: my final approval (09-17) dropped a caveat my first pass (09-16) had stated clearly — that actual email delivery on accept-and-resend was never verified. Restating the testing picture explicitly, split by source, per Lui's feedback on this pattern.

**Decision:** Approved stands at the code level — no unresolved review finding across either pass (concurrency fix verified correct, injection/XSS checks clean, both of Lui's flagged questions — fail-open rights gate, exit()-in-shared-mail-path — traced and resolved).

**Testing, split explicitly:**
- **I ran myself:** `php tests/test_email_lookalike.php` — 21/21 pass, matches the PR's claim. Read the full 391-line lookalike table for coverage. Manually checked header injection, SQLi, and XSS paths against current code — all clean.
- **Author-reported, not independently reproduced by me:** the manual ssl12 order-3342 walkthrough (lookalike detected, suggestion shown, accept button updates `ordrer.email`) — I did not personally drive this in a browser.
- **Never verified by anyone:** actual outbound delivery after accept-and-resend. The PR body says so directly — ssl12 has no outbound mail, and customer confirmation of a successful resend on the live incident (saldi_390 order 162742, manually patched 2026-09-09) is still pending.

**Does the unverified delivery check block merge or deployment?** Not merge — there's no environment available pre-merge to test real outbound delivery, and the write path itself (DB update to `ordrer.email`) is independently verified safe regardless of delivery outcome. It does need confirming **before this is treated as resolved in production**: an actual successful resend needs to be confirmed for order 162742, and ideally one live accept-and-resend send observed post-deploy. **Owner: PR author** (or whoever holds SST-759) — get that confirmation before closing the ticket, not before merging.

**Housekeeping, unrelated to review content:** this PR currently shows `mergeable: CONFLICTING` against master — needs a rebase/merge before it can go in regardless of review status.

**Nonblocking:** fail-open rights-gate shape matches the existing convention in includes/online.php — FYI, not a new hole.

debitor/formularprint.php
```diff
@@ -31,6 +31,7 @@
 // 20260102 LOE Added department support for background files
 // 20260309 PHR Fixed error in $returside after printing
 // 20260309 PHR Fixed another error in $returside after printing
+// 20260909 Sawaneh SST-759: POST email_fix_from applies the recipient suggestion from send_mails() after explicit acceptance
```
ZaynSaul, line 34: History-comment collision with master's JOB-124 lines from #605 — trivial, keep both.

debitor/formularprint.php
```diff
@@ -89,6 +90,26 @@ function find_background_file($background, $file_type, $department) {
 
     return null;
 }
+// 20260909 Sawaneh SST-759: the user accepted the suggested recipient correction shown by
+// send_mails(); the suggestion is re-derived server-side from the rejected address.
+if (isset($_POST['email_fix_from']) && isset($_POST['id'])) {
```
ZaynSaul, line 97: Real conflict, not just a diff overlap: master (#605) rewrote this block to sanitize $returside via nav_sanitize_returside() and added the $isPopupRequest flag; this branch inserts the email_fix_from handling and its own plain $returside = if_isset(...) at the same spot. Both blocks are needed — when resolving, keep master's sanitized $returside/$isPopupRequest logic and add this branch's email_fix_from block ahead of it, not instead of it.

importfiler/tekster.csv
```diff
@@ -5098,4 +5098,16 @@
 5151	Eksporter CSV (alle sider)	Export CSV (all pages)	Eksporter CSV (alle sider)
 5152	Eksporter alle konti under de valgte filtre til CSV, uden sideopdeling	Export all accounts under the selected filters to CSV, without pagination	Eksporter alle kontoer under de valgte filtrene til CSV, uten sideinndeling
 
+5210	Ugyldig e-mailadresse	Invalid e-mail address	Ugyldig e-postadresse
```
ZaynSaul, line 5102: Text-ID collision, not just a merge conflict: this reserves 5210-5221, but master already has 5231 (Udført af) right after this branch's 5230 line. Please confirm 5210-5221 are still unclaimed against current master before resolving — if anything else grabbed that range in the meantime this needs renumbering, not a textual merge.

ZaynSaul (changes requested):
This branch now conflicts with `master` (GitHub shows `CONFLICTING` / `DIRTY`) after #605 (SST-789, "Print all selected invoices in one ordered PDF") merged, which also touched `formularprint.php`'s request-handling section.

Checked with `git merge-tree` — two files actually conflict, and unlike a simple rebase these need real manual merging, not just "keep both":

1. **`debitor/formularprint.php`, top of file** — history-comment collision (trivial): this branch's SST-759 line vs master's JOB-124 `returside`/popup lines from #605. Keep both.
2. **`debitor/formularprint.php`, right after the `nav_popup_query()` block** — this is the real one. Master (#605) rewrote the `$returside` handling (added `nav_sanitize_returside()` and the `$isPopupRequest` flag from the GET/POST popup fix), while this branch inserts the `email_fix_from` handling and its own `$returside = if_isset($_POST['returside'])` at the exact same spot. These aren't alternatives — both blocks are needed, and whoever resolves this needs to check that `$returside` ends up sanitized (master's version) rather than reverting to this branch's plain `if_isset()`, and that the popup-flag logic from #605 isn't dropped.
3. **`importfiler/tekster.csv`** — an actual text-ID collision, not just a diff conflict: this branch claims IDs `5210`-`5221` for the new lookalike-feedback strings, but master already has `5231` (`Udført af`, from #630/#639) sitting right after this branch's `5230 Mobilnummer` line, and — more importantly — text IDs are a shared namespace, so if anything else has claimed 5210-5221 in the meantime this needs a fresh unused range, not just a textual merge. Please confirm 5210-5221 are still free against current master and renumber if not.

Requesting changes to get this resolved properly — the design/logic itself was already approved (mine + CodeRabbit); this is scoped to the conflict only. Given the overlap in `formularprint.php` touches both PRs' actual runtime logic, I'd suggest whoever resolves it (Sawaneh or reviewer) re-test the popup-close and email-fix flows after merging, not just eyeball the diff.

ZaynSaul (approved):
Verified the merge (51f63b37) against my three conflict items from the earlier review:

1. Header comment - both SST-759 and JOB-124 lines kept.
2. The real logic conflict - $returside correctly uses nav_sanitize_returside() from #605, not reverted to plain if_isset(), $isPopupRequest is intact and still used, and this branch's email_fix_from handling is present alongside it. Nothing dropped from either side.
3. tekster.csv - checked every ID this PR's files actually reference (5210-5221 plus the pre-existing 2172/3370) against the current merged file: no duplicates anywhere.

Also lint-checked the merged formularprint.php (clean) and ran test_email_lookalike.php on the actual merged head in a worktree: 21/21 pass.

Approved - conflict resolution is correct, design/logic was already approved earlier. The one still-open item from my first approval stands: the accept-and-resend flow itself needs a live end-to-end check before this ships, since ssl12 has no outbound mail to verify delivery against.

PR #547, opened 2026-09-03
Add local OIOUBL and Peppol invoice extraction

PeterRude:
## Summary

- accept XML files in the document-pool upload flow and preserve their `.xml` extension
- parse OIOUBL/Peppol Invoice and CreditNote documents locally without an AI request
- populate amount, date, supplier, invoice number, description, and currency through the existing metadata flow
- reject XML documents containing DOCTYPE or entity declarations

## Validation

- `php -l includes/documents.php`
- `php -l includes/docsIncludes/docPool.php`
- `php -l includes/docsIncludes/invoiceExtractionApi.php`
- `php tests/test_invoice_extraction_api.php` — 16 passed
- `php tests/test_doc_pool_upload_exactly_once.php` — 11 passed
- `git diff --check`

## Manual testing

- upload an OIOUBL invoice and confirm date, amount, supplier, invoice number, and currency
- upload a Peppol BIS 3 invoice and confirm the same metadata
- scan an existing XML pool file using the extraction button
- test a UBL CreditNote
- upload PDF/JPG/PNG and confirm they still use the existing AI extraction flow

## Summary by CodeRabbit

* **New Features**
  * Added support for uploading OIOUBL/Peppol XML invoices alongside PDF and image files.
  * XML invoices are validated and processed locally, extracting key invoice details such as number, date, vendor, amount, currency, and description.
  * Added protection against unsafe XML declarations during invoice processing.

* **Bug Fixes**
  * Improved file handling so uploaded documents retain the appropriate file extension and processing method.

PR #588, opened 2026-09-10
Dokumenter: Tilføj manglende XML-upload og lokalt UBL-udtræk

PeterRude:
XML-fakturaer kunne udtrækkes i den lokale installation, men understøttelsen manglede på master: uploadvælgeren accepterede ikke XML, uploadforløbet gemte altid som PDF, og extractInvoiceData havde ingen lokal UBL-parser.

Tilføjer den lokale OIOUBL/Peppol UBL-udtrækning af beløb, dato, leverandør, fakturanummer, beskrivelse og valuta. XML behandles før opslag af API-nøgle og sendes ikke til AI-tjenesten. Invoice og CreditNote understøttes; fejlformet XML samt DOCTYPE/entity-deklarationer afvises. Dette er feltudtrækning, ikke fuld UBL-skemavalidering.

XML tillades i filvælger og drag/drop samt begge uploadforløb. Filtypen bevares i lagring, omdøbning, metadata og uploadsvaret. Integreret med masters nyere FileReservation, så beskyttelsen mod overskrivning ved navnesammenfald bevares, også mellem XML og PDF.

## Validering
- `php tests/test_invoice_extraction_api.php`: 21 bestået. Omfatter XML uden API-nøgle, faktura/kreditnota, fallback for leverandørnavn og valuta, ugyldig XML, navnekollision samt eksisterende PDF/API-forløb.
- `php tests/test_doc_pool_upload_exactly_once.php`: 11 bestået.
- PHP-syntakskontrol for alle fire ændrede filer og `git diff --check` bestået.
- Ingen ekstern AI-tjeneste eller databaseændringer brugt i testene. Browserforløbene nedenfor er ikke gennemført.

## Manuel kontrol ved review
- `includes/docsIncludes/docPool.php`: Upload samme UBL-fil via filvælger og drag/drop. Kontrollér, at XML accepteres, og at metadata kun gemmes én gang.
- `includes/documents.php`: Kontrollér, at filen fortsat hedder `.xml`, og at beløb, dato, fakturanummer, beskrivelse og valuta følger med. Genåbn dokumentet fra puljen. Gentag med to filer med samme navn og med samme leverandør/dato; begge dokumenter skal bevares med forskellige navne.
- `includes/docsIncludes/invoiceExtractionApi.php`: Upload en UBL-faktura uden API-nøgle med automatisk udtræk slået til. Kontrollér felterne, og gentag med automatisk udtræk slået fra. Kontrollér også PDF/JPG-upload med normal API-konfiguration.
- Kontrollér både den moderne pulje/AJAX-upload og det ældre uploadforløb fra en bilagstilknytning.

- [ ] I have read and understood the developer guidelines https://docs.google.com/document/d/1GOmomtvKf21OV2VWNOIPweDi4gJHpMCK5qXzrPrypUc/edit?usp=sharing

### DIFF
--- a/includes/docsIncludes/docPool.php
+++ b/includes/docsIncludes/docPool.php
@@ -43,6 +43,7 @@
 //                 which the very next INSERT/UPDATE in each of those code paths already
 //                 references - a brand-new tenant's first pool file would fail with
 //                 "column norm_amount does not exist". Added the column to both.
+// 20260910 CDX/PHR Enable local UBL XML invoice upload and extraction.
 include_once(__DIR__ . "/poolAmountNormalizer.php");
 /**
  * Log message to a file in temp/$db/docPool.log
@@ -3836,7 +3837,7 @@ JS;
 	print "<div style='padding: 12px;'>";
 	
 	// Unified upload zone (click to select or drag and drop)
-	print "<input id='fileUploadInput' type='file' name='uploadedFile[]' accept='.pdf,.jpg,.jpeg,.png' multiple style='display:none'>";
+	print "<input id='fileUploadInput' type='file' name='uploadedFile[]' accept='.pdf,.jpg,.jpeg,.png,.xml' multiple style='display:none'>";
 	print "<div id='dropZone' ondrop='handleDrop(event)' ondragover='handleDragOver(event)' onclick='document.getElementById(\"fileUploadInput\").click()' style='width: 100%; border: 2px dashed #bbb; border-radius: 10px; padding: 90px 16px; background-color: #f8f8f8; cursor: pointer; transition: all 0.3s ease; box-sizing: border-box; display: flex; align-items: center; justify-content: center; margin-bottom: 12px;'>";
 	print "<div id='dropText' style='display: flex; flex-direction: column; align-items: center; gap: 8px; pointer-events: none; text-align: center;'>";
 	print "<svg viewBox='0 0 24 24' fill='none' stroke='#7ab3d4' stroke-width='1.5' width='44' height='44'><path d='M14 2H6a2 2 0 0 0-2 2v16a2 2 0 0 0 2 2h12a2 2 0 0 0 2-2V8z'/><polyline points='14 2 14 8 20 8'/><line x1='16' y1='13' x2='8' y2='13'/><line x1='16' y1='17' x2='8' y2='17'/></svg>";
@@ -3874,7 +3875,7 @@ JS;
 	}
 
 	function uploadFiles(files) {
-		var allowedExtensions = ['.pdf', '.jpg', '.jpeg', '.png'];
+		var allowedExtensions = ['.pdf', '.jpg', '.jpeg', '.png', '.xml'];
 		var validFiles = [];
 		for (var i = 0; i < files.length; i++) {
 			var fileName = files[i].name.toLowerCase();
--- a/includes/docsIncludes/invoiceExtractionApi.php
+++ b/includes/docsIncludes/invoiceExtractionApi.php
@@ -2,6 +2,7 @@
 // --- includes/docsIncludes/invoiceExtractionApi.php -----
 // Helper functions for invoice extraction API integration
 
+// 20260910 CDX/PHR Enable local UBL XML invoice upload and extraction.
 function invoiceExtractionApiResolveApiKey() {
 	if (!function_exists('db_select') || !function_exists('db_fetch_array')) {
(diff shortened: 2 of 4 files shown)


LuiHedlund02:
Lead reviewer: @ZaynSaul
Reason: XML upload path + local UBL parsing in shared document code; #547 is the older version of the same change and will be closed.
Focus: MIME check loosened to MIME-or-extension; how the pool viewer serves a stored `.xml` back (inline text/html would be stored XSS); XXE defence is a `stripos` scan for DOCTYPE/ENTITY plus `LIBXML_NONET`.
@SirRolin one question: `PayableAmount` goes through `normalizePoolAmount()`, whose three-trailing-digits heuristic turns a UBL `1250.500` into `1250500`. Should XML amounts bypass it and parse as strict decimal?
Verification: 21 assertions incl. XXE and CreditNote; browser flows not run.
Next action: @PeterRude rebase; review on the diff now.

LuiHedlund02:
@PeterRude resolve conflicts before @ZaynSaul’s extraction review. Focus on preserving FileReservation upload behavior, XML parsing restrictions and serving XML with a safe content type/disposition. @SirRolin targeted question: should UBL PayableAmount bypass normalizePoolAmount(), which can turn 1250.500 into 1250500? Please establish the expected decimal interpretation. #547 is proposed for closure in favor of this PR.

ZaynSaul (dismissed):
Reviewed the XML upload / local UBL extraction feature. Path traversal and upload-validation are clean - basename()+sanitize_filename() matches the safe pattern already established from the debitor/peppol.php fix, and the MIME/extension fallback just makes the AJAX and legacy upload paths consistent with each other, not a new weakening.

One finding worth a comment, not blocking: the XXE guard in invoiceExtractionApi.php does a raw-byte stripos() for `<!DOCTYPE`/`<!ENTITY` before parsing. I tested this empirically with a UTF-16-encoded payload (BOM + DOCTYPE/ENTITY) - the textual filter is bypassed since those ASCII substrings aren't contiguous in UTF-16 bytes, and DOMDocument::loadXML() auto-detects the encoding and parses it anyway. However, the entity itself does NOT get expanded in this environment (PHP/libxml disables external SYSTEM entity resolution by default, and LIBXML_NOENT isn't passed), so this isn't currently exploitable - it's fragile defense-in-depth relying on implicit parser defaults rather than an explicit guarantee. Suggest also checking `$document->doctype === null` post-parse as a belt-and-suspenders measure alongside the existing string scan.

Good to see you already tested the classic file:///etc/passwd XXE payload in the test suite. Approved - the manual browser-flow checklist in the PR body (upload picker/drag-drop, duplicate-name collision, auto-extract on/off) still needs a live pass per your own notes.

LuiHedlund02:

ZaynSaul (changes requested):
Confirmed XML is dispatched before any API/network code runs, and the FileReservation cross-extension collision protection works as claimed (ran the test suite on the branch, 21/21 pass).

Requesting changes on two things:

1. **The DOCTYPE/ENTITY rejection doesn't do what it claims.** Checked includes/docsIncludes/invoiceExtractionApi.php:75 directly — it's `stripos($xmlContent, '<!DOCTYPE') !== false || stripos($xmlContent, '<!ENTITY') !== false`, a raw-byte substring scan on the content BEFORE parsing. That's encoding-dependent: a non-UTF-8-declared payload (e.g. UTF-16LE with BOM) doesn't have those ASCII bytes contiguous, so the check misses it while DOMDocument::loadXML() still parses it fine afterward. The PR body explicitly claims malformed XML and DOCTYPE/entity declarations are rejected — that's only true for UTF-8 input. Fix: check $document->doctype !== null after parsing (encoding-agnostic) instead of, or in addition to, the pre-parse string scan.

2. **PayableAmount is a structured xsd:decimal but gets piped through normalizePoolAmount()'s ambiguous-separator heuristic**, which treats a single dot followed by exactly 3 digits as thousands-grouping rather than a decimal. A perfectly valid UBL amount like "1250.500" becomes 1250500 — a 1000x error. Lui already asked this exact question on the PR thread and it's still unanswered in the current diff. XML amounts are unambiguous and shouldn't go through a heuristic built for OCR/AI free text.

Non-blocking: docPool.php's existing XML viewer writes EasyUBL's HTML response to a temp file unsanitized and frames it — pre-existing, not this PR's code, but this PR does add a second, more directly user-controlled route into that same viewer via ordinary doc-pool upload. Worth a follow-up ticket.

PR #591, opened 2026-09-11
SD-686: grid filter defaults are honoured

Loe4g:

A grid filter option can be declared active from the start with `"checked" => "checked"` — or with a value computed from page state, e.g. `$skjul_lukkede ? "" : "checked"` — but that default never took effect: on a user's first view the checkbox was always empty, and the filter only worked after the user ticked it and saved.

**Cause.** `fetch_grid_setup()` wrote the filter *definitions* (`json_encode($filters)`) into `datatables.filter_setup`, while `updateCheckedValues()` — the only reader of that column — expects the saved *selection* map (`{"Misc":{"Vis udgået":"checked"}}`) that `save_filter_setup()` writes. A definition list has no `$secondArray[$filterName]` key, so the else-branch forced every option back to `''`; that also flipped `build_query()` to the `sqlOff` branch (`includes/grid.php:709-714`).

**Fix.**
- `fetch_grid_setup()` stores an empty selection map (`'{}'`) instead of the definitions — definitions are not selections.
- `updateCheckedValues()` lets a saved selection win per option, and where nothing has been saved it now honours the declared default; an option with no declared `checked` key resolves to `''` (no new undefined-key warnings).
- `render_filters()` gives every checkbox a hidden companion input, so an **unticked** option is submitted too. An unchecked box is not posted at all, so without this the panel stored an empty map — indistinguishable from "never saved" — and a declared default could never be turned off once saved.
- Applied to all four grid implementations: `includes/grid.php`, `includes/kreditorOrderFuncIncludes/creditor_orderlist_grid.php`, `includes/orderFuncIncludes/grid_order.php`, `includes/orderFuncIncludes/grid_account_lookup.php`.

**Behaviour consequence (intended).** Any page that declares a checked default now honours it on the first view: `lager/lister/{ordrestatus,indkøb,vareliste}` ("Vis udgået" ticked ⇒ discontinued items shown), `debitor/debitor.php` (`$skjul_lukkede`), `debitor/ordreliste.php` (`$valg`). A user's own saved selection always wins — including an explicit untick, which now persists across reloads.

## How I verified this

**Reproduced on master** (`c06274da`, tenant `tenant_db`, user `username`, grid `ordrestat`): with no `datatables` row for that user + grid, opening Lager → Ordrevisning → Redigér → Filtre showed "Vis udgået" **unchecked**, although `lager/lister/ordrestatus.php:491` declares `"checked" => "checked"`; the list hid discontinued items. The row that page load created held the definition list (`[{"filterName":"Misc",…"checked":"checked"…}]`), the format the reader cannot use (visible in the application's modify/query log).

**After the fix:** the same first view renders the option **checked**, and the row holds `{}`.

**Automated test:** new `tests/characterization/includes/UpdateCheckedValuesTest.test.php` — 13 assertions covering declared default (fresh `{}` / legacy definition-list row / NULL / empty), declared `""`, saved-selection precedence (ticked and unticked), the hidden companion input that makes an untick persist, an option with no declared `checked` key, and a warning trap. Result here: `OK, Tests: 13, Assertions: 14` (PHPUnit 10.5 phar on PHP 8.1).
**`composer test` was not run on this machine:** `composer.lock` pins PHPUnit 11.5.56, which requires PHP ≥ 8.2, and this box only has PHP 8.1 (no `vendor/`). The new test is picked up by the `characterization` suite in CI (`phpunit.xml`, `.test.php` discovery).

**Manual scenarios per affected file/function (pre-PR test report):**

| File / grid | Scenario |
|---|---|
| `lager/lister/ordrestatus.php` (`ordrestat`) | Remove the user's `datatables` row → open the filter panel: "Vis udgået" ticked, discontinued items (`lukket='1'`) listed. |
| `lager/lister/ordrestatus.php` (`ordrestat`) | Untick "Vis udgået" → Gem → reload the list: stays unticked and discontinued items stay hidden; `filter_setup` = `{"Misc":{"Vis udgået":""}}`. Also try "Fjern alle" → Gem → reload: stays off. |
| `lager/lister/indkøb.php` (`indkøb`), `lager/lister/vareliste.php` (`varelst1`) | Same first-visit check (tick is honoured on a fresh row). |
| `debitor/debitor.php` (`debitor_list`) | View setup that does *not* hide closed debtors → Misc "Vis udgået" ticked, closed debtors listed; with `$skjul_lukkede` set → unticked. |
| `debitor/ordreliste.php` (`ordrelst_<valg>`) | `?valg=ordrer` / `faktura` / `tilbud` / `pbs` → Redigér → Filtre: the matching Ordretype checkbox reflects the mode. |
| `kreditor/productLookup.php` | Filter panel: Base/Find/Supplier defaults ticked. Results unchanged — the same conditions are hard-coded in the page's `$where_conditions`, so they were already applied. |
| Regression | Users with saved filters: selections still win. Grids with no declared default: unchanged. |

**Note for existing saved setups:** rows saved before this change contain only *ticked* options. For such a row, an option the user had unticked was never recorded, so it will show the declared default once — saving the panel again (now with the hidden companion inputs) records it explicitly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Grid filters now honor their configured default selections on first load.
  - Saved filter choices correctly override defaults when present.
  - Clearing a default filter is now preserved after saving.
  - Unselected filter options are submitted reliably.

- **Tests**
  - Added coverage for default handling, saved selections, cleared filters, and warning-free filter processing.

### DIFF
--- a/includes/grid.php
+++ b/includes/grid.php
@@ -22,6 +22,7 @@ Regards:) 20260220 LOE
 // 20260817 Sawaneh Sort descending columns NULLS LAST, honor defaultSortDirection on
 //                  first header click and validate the request-sourced sort value.
 // 20260910 CDX/PHR Preserve a literal zero search in both row and count queries.
+// 20260911 LOE SD-686: grid filter defaults declared with "checked" are honoured.
 ######################### >>>>>>>EndNotice<<<<<<<<<<<<##############################
 /**
  * Extracts values from a specific column in a multi-dimensional array.
@@ -560,7 +561,11 @@ function fetch_grid_setup($id, $columns_filtered, $search_setup, $filters) {
 
         // Encode configurations as JSON for storage
         $columns_json = db_escape_string(json_encode($columns_save));
-        $filters_json = db_escape_string(json_encode($filters));
+        // SD-686: a fresh grid row starts with an empty *selection* map. Filter
+        // defaults belong to the page's $filters definition and are applied by
+        // updateCheckedValues(); storing the definitions here left the declared
+        // defaults unreadable on the first page view.
+        $filters_json = '{}';
         $search_json  = db_escape_string(json_encode($search_setup));
 
         // Insert the new grid setup into the database
@@ -621,16 +626,18 @@ function fill_missing_values($firstArray, $secondArray) {
  * @return array The updated first array with the 'checked' values for options updated.
  */
 function updateCheckedValues(array $firstArray, array $secondArray) {
+    // SD-686: only a saved *selection* map may override a filter option's declared
+    // default. Where nothing (or only a legacy definition list) has been saved, the
+    // declared value stays in force instead of being forced back to ''.
     foreach ($firstArray as &$filter) {
         $filterName = $filter['filterName'];
-        if (isset($secondArray[$filterName])) {
-            $updatesForFilter = $secondArray[$filterName];
-            foreach ($filter['options'] as &$option) {
-                $option['checked'] = isset($updatesForFilter[$option['name']]) ? $updatesForFilter[$option['name']] : '';
-            }
-        } else {
-            foreach ($filter['options'] as &$option) {
+        $updatesForFilter = isset($secondArray[$filterName]) ? $secondArray[$filterName] : array();
+        foreach ($filter['options'] as &$option) {
+            if (!isset($option['checked'])) {
                 $option['checked'] = '';
+            }
+            if (isset($updatesForFilter[$option['name']])) {
+                $option['checked'] = $updatesForFilter[$option['name']];
             }
         }
     }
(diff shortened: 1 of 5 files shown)


LuiHedlund02:
@Loe4g proposing closure because #597 contains these filter-default changes. Please confirm or identify a remaining difference that needs preserving. I will leave this open for agreement, or close after 2026-09-23 if there is no objection.

ZaynSaul (approved):
Reviewed SD-686. This is a real bug fix, not a default-value change - traced the old updateCheckedValues() against current code and confirmed the root cause: it conflated filter *definitions* (a numeric-indexed JSON array) with filter *selections* (a name-keyed map), force-zeroing a freshly-created row's real definitions whenever no saved selection existed yet.

Verified the fix correctly preserves declared defaults unless explicitly overridden, correctly stores '{}' instead of the definitions blob on first-ever row creation, and correctly persists an explicit clear (empty-string selection). The hidden value='' input ensures an unchecked box still submits something, and I traced that against save_filter_setup()'s existing on->checked normalization - sound.

One thing worth flagging before merge: this PR and #597 (SD-685) touch the exact same 4 files/functions and both add an identically-named characterization test file - #597 appears to build directly on this one. Recommend merging this PR first, then rebasing #597 on top, to avoid a near-certain conflict.

Approved.

Loe4g:
All changes now reside in #597

PR #597, opened 2026-09-11
Sd 685 grid setup translation

Loe4g:

The grid stored **display text as data**: filter selections were keyed by the filter/option labels, and `headerName`/`description` were persisted into `datatables.column_setup` where they outranked the code. A translated label therefore detached a stored selection from its checkbox, headers froze in the language a list was first opened in, columns added to the code never reached existing users, and a user-added column could only be named by hand.

**Filters** — `updateCheckedValues()`/`render_filters()` look a stored selection up by a declared `filterKey`/`optionKey`, with the display text as fallback so rows saved before keys existed keep working. Keys added for `lager/lister/{ordrestatus,indkøb,vareliste,serialnumber}.php` (DB-sourced option labels keyed by id: `vg_<kodenr>`, `lev_<kontonr>`) and `debitor/ordreliste.php`.

**Columns** — `fill_missing_values()` → `merge_column_setup()`: the code defines which columns exist and what they are called; the stored row contributes preferences only (order, width, align, visibility, optional `customHeaderName`/`customDescription`) matched on the language-independent `field`. `save_column_setup()` keeps a row by `field` (an empty header no longer discards the column) and records `visible=false` for a removed column; the editor shows the code's text as `placeholder` and receives the merged list.

**A header saved before `customHeaderName` existed is no longer dropped.** It is kept as `customHeaderName` unless the code provably produces that text — every stored translation of the column's `tekst_id`, plus the default after the `|`. A personal rename therefore survives, while a header frozen in the language it was saved in is still discarded. Columns whose header comes from translations declare `headerText` (the 17 `findtekst(...)` headers in `debitor/ordreliste.php`), which is what makes the comparison exact; a literal or tenant-built header (e.g. the warehouse columns built from `grupper`) has no set of code forms, so its stored value is kept rather than dropped.

Applied to all four `create_datagrid()` implementations: `includes/grid.php`, `includes/orderFuncIncludes/grid_order.php`, `includes/orderFuncIncludes/grid_account_lookup.php` (filters + merge only — no column editor), `includes/kreditorOrderFuncIncludes/creditor_orderlist_grid.php`.

**Base and merge order:** based directly on `master` (`a5280ddf`). SD-686 (`435dd053`, "grid filter defaults are honoured") is the first commit in this branch, so its behaviour is included here; PR #591 is proposed for closure as contained, per Lui. The keyed lookup relies on SD-686's rule that a declared default applies when nothing is saved, and on its hidden companion input that makes an unticked option persist.

**Consequences (accepted — worth a release-note line)**
- Existing filter selections keep working, and a language switch no longer detaches them.
- Columns: a legacy user whose curated list omitted code columns sees them once, until they save. A historical personal rename survives (it is re-read as `customHeaderName`); renames saved from now on are stored under `customHeaderName` and are language-independent.
- A stored header that happens to be an *older* code text (a literal changed since, or a renamed warehouse) is kept as the user's own and stays visible until they clear it in the editor — the deliberate side of keeping renames rather than discarding them.
- `finans/kladdeliste.php`'s private filter format is tolerated; only its panel checkbox changes (cosmetic).

**Out of scope:** the seven pages that declare filters but no keys — `debitor/debitor.php`, `debitor/debitor_kommission.php`, `debitor/debitor_historik.php`, `finans/kontospec.php`, `finans/kladdeliste.php`, `kreditor/kreditor.php`, `kreditor/productLookup.php` — keep today's behaviour until they adopt keys; no migration needed. The Lager files' translated `findtekst(...)` lines stay commented out (translation task).

## Review follow-ups (Lui's handoff)

1. **Escaping.** `save_column_setup()`/`save_filter_setup()` escaped `tabel_id` in the SELECT this PR added but interpolated `$id` raw in the UPDATE beside it; both now use `db_escape_string()` (`tabel_id` is a string identifier such as `'ordreliste'` or `"productLookup$id"`, so not `intval()`).
2. **Array guards — not reproducible.** `includes/grid.php` already normalises every consumed setup through `decode_grid_json_array()`, which handles `is_array`, `null`, `''` and malformed JSON with a per-call-site fallback — the same guard the other three copies apply inline. No change made.
3. **Preserved renames** — implemented as described above.

CodeRabbit's review on the current head is APPROVED. The two findings from its earlier rounds (legacy-renderer coverage, `tempnam()` leak) were fixed in `0c5d0ba`/`2c1ae2a` and confirmed resolved.

## How I verified this

Behaviour compared on live data across **three code states** — pre-PR master, the PR as first reviewed, and the current head — by seeding a stored setup and reading the header the grid actually renders:

| stored shape | pre-PR | as reviewed | current head |
|---|---|---|---|
| literal / tenant-built column, stored abbreviation | shown | code text | **shown** |
| legacy `headerName` = a personal rename | shown | **dropped** | **shown** |
| legacy `headerName` = a text the code also produces (a translation) | shown — **frozen** | code text | **code text** |
| `customHeaderName` = a personal rename | ignored | shown | **shown** |

Relative to pre-PR nothing that used to be displayed is lost; the only two differences are the intended ones — a header frozen in another language now follows the session language, and a rename saved under `customHeaderName` is honoured.

- **Suites:** the three grid suites are green — `GridSetupTranslationTest` (14 tests, 26 assertions), `SaveColumnSetupTest` (18/48), `UpdateCheckedValuesTest` (13/14); 45 tests / 88 assertions together. `GridSetupTranslationTest` pins the keyed lookup through a translation + language switch, legacy text-keyed rows, translated header *and* sub-header reaching a user who saved in another language, user order/width, a new code column appearing, a removed column staying hidden, a code-hidden column only when surfaced, custom header wins / empty falls back, a rename the code cannot produce being kept, and an older code text being discarded.
- **Whole `characterization` suite, both states, same command:** master `228 tests, 1 error, 7 failures`; this branch `273 tests, 1 error, 7 failures` — the **same** error and the same seven failures, so the delta is +45 green and nothing new broken. Those eight are pre-existing and environmental (`UsdecimalCharacterizationTest` fails only when the suite runs as one process, and `NewlabelCharacterizationTest` can't reach a tenant database from the CLI); both reproduce on master.
- **Runtime:** `composer test` cannot run in my environment (PHPUnit 11 requires PHP >= 8.2, this box has 8.1.2), so the suite was run with PHPUnit 10.5, which supports the attributes these tests use. No new PHP warnings on touched lines; the deprecation count rises from 1 to 5 only because the new tests load `includes/grid.php`, whose pre-existing `build_query()` signature is deprecated on PHP 8.
- **Live pages**, one per framework copy, all HTTP 200 with no fatals/warnings/notices: `lager/lagerstatus.php`, `debitor/ordreliste.php?valg=ordrer`, `kreditor/ordreliste.php?valg=ordrer`, `debitor/ordre.php`.

**Manual scenarios** (tenant `tenant_db`, which has English translations; language switched with `update brugere set language_id = 2`)
1. **Filters** (`updateCheckedValues`, `render_filters`): set a filter on Ordreliste → switch language → reload: the same option is still ticked, and `filter_setup` holds keys (`{"ordretype":{"ordrer":"checked"}}`).
2. **Column headers** (`merge_column_setup`): open a Lager list saved in Danish → switch language → headers and sub-headers follow the translation; `Redigér → Kolonner` shows the code's text as placeholder.
3. **Column list and editor** (`save_column_setup`, `render_columns`): a column added in the code appears for an existing user; a deleted column stays hidden after reload; a custom header wins and clearing it restores the code's text.
4. **Lager first visit / untick** (`ordrestat`): fresh user → "Vis udgået" ticked, discontinued items listed; untick + Gem → stays unticked.
5. **Regression and consumers**: saved filters still win; a page without keys behaves exactly as before; `debitor/ordreliste.php` reads only `field` from `column_setup` (compatible) and `finans/kladdeliste.php` loads and remembers its filter.

**Pre-existing, untouched:** `build_query()`/`build_count_query()` in `includes/grid.php` declare an optional `$searchTerms` before required `$sort`/`$offset` (PHP 8 deprecation) — separate ticket.


Loe4g:
@coderabbitai review

Loe4g:
@coderabbitai review

LuiHedlund02:
Lead reviewer: @ZaynSaul
Reason: rewrites grid setup resolution in all four framework copies; the diff contains all of #591.
Focus: `merge_column_setup()` drops stored personal column renames (`headerName` -> `customHeaderName` has no migration), Lui decides whether a one-time copy is needed; `includes/grid.php` lacks the `is_array` guards the other three copies got; `save_column_setup()` UPDATE still interpolates `tabel_id='$id'` raw.
Verification: three characterization test files (632 lines), not run by CI.
Next action: @Loe4g resolve conflicts; #591 is redundant and can close; #582 lands after this. Review on the diff now.

LuiHedlund02:
@Loe4g please resolve conflicts and preserve stored personal column renames when moving headerName to customHeaderName (migrate/read the old stored field rather than dropping user customization). Apply the array guards consistently across the grid copies. @ZaynSaul remains lead after those fixes. #591 is proposed for closure as contained here; #582 should follow this PR.

ZaynSaul (dismissed):
Reviewed SD-685 - despite the title, this isn't new translation strings, it's a real bug: the grid framework was persisting display *text* (option labels, headerName/description) as data, so a language switch detached saved filter selections from their checkbox and froze column headers in whichever language they were first saved in. The filterKey/optionKey/field stable-identifier fix correctly addresses that root cause.

All 6 CodeRabbit findings across 3 review rounds (2 critical) verified fixed against the current head - the stale-merged-column-list bug, the empty-column_setup-on-every-save bug, and the visible-row-wrongly-restored-on-field-change bug are all confirmed resolved. CodeRabbit's own final review at the current head is APPROVED, consistent with my read.

As noted on #591: this PR builds directly on that one (same 4 files) - please merge #591 first, then rebase this on top, since GitHub is already showing CONFLICTING against #591 as expected.

Approved, contingent on the #591-then-#597 merge order. A live check of filter/column persistence across an actual language switch is worth doing as a follow-up.

Loe4g:
@ZaynSaul rebased onto `master` (`a5280ddf`) and pushed `2397c15e` — the rewrite dismissed your approval, so it needs a fresh look. Three commits since your review, all from Lui's handoff:

1. **Escaping (Lui 3).** `save_column_setup()`/`save_filter_setup()` escaped `tabel_id` in the SELECT this PR added but interpolated `$id` raw in the UPDATE beside it. Both now use `db_escape_string()` — `tabel_id` is a string identifier (`'ordreliste'`, `"productLookup$id"`), so not `intval()`.
2. **Preserved renames (Lui 1).** A header saved under the old `headerName` key was dropped outright. It is now kept as `customHeaderName` **unless the code provably produces that text** — every stored translation of the column's `tekst_id`, plus the default after the `|`. A personal rename survives; a header frozen in the language it was saved in is still discarded. The 17 translated headers in `debitor/ordreliste.php` now declare `headerText`, which makes the comparison exact; literal or tenant-built headers (e.g. the warehouse columns) have no form set, so their stored value is kept rather than dropped.
3. **Array guards (Lui 2) — not reproducible.** `includes/grid.php` already normalises every consumed setup through `decode_grid_json_array()`, handling `is_array`, `null`, `''` and malformed JSON with a per-call-site fallback — the same guard `grid_order.php` does inline. No change made.

Evidence is in the description: behaviour compared across pre-PR master / as you reviewed / current head on live data, and the suite comparison (master 228 tests vs 273 here, identical pre-existing failures, all 45 grid tests green). Nothing outside those three commits moved. CodeRabbit's review on this head is APPROVED.

One note: this is now based directly on master, so the earlier "merge #591 first" requirement no longer applies — SD-686's commit rides inside this branch and #591 is proposed for closure as contained.

ZaynSaul (approved):
Verified the central design question directly: betweenUpdates.php is untouched, so this is a deliberate read-side fallback rather than a data migration. Confirmed merge_column_setup() still keys on 'field' (unchanged, already the primary key pre-PR) and updateCheckedValues() tries the new stable key first with a fallback to the old label — checked this line-by-line in grid.php. grid_known_header_texts() correctly distinguishes a genuine personal rename from a frozen stale translation by checking against every producible tekst_id/sprog_id combination. Verified the same merge/preservation/escaping logic is applied consistently across all 4 create_datagrid() copies (grid.php, grid_order.php, grid_account_lookup.php, creditor_orderlist_grid.php), and that Lui's three 09-16 handoff findings (legacy headerName loss, missing is_array guards, raw tabel_id interpolation) are all addressed in current head.

Two accuracy notes, non-blocking:
1. The PR body's claim that 'a language switch no longer detaches' existing filter selections is a slight overstatement — a row saved in language A and viewed in language B for the first time after this ships still won't match on that first view (self-heals on next save). Worth a one-line clarification.
2. The body's reference to fixes in commits '0c5d0ba'/'2c1ae2a' for a legacy-renderer/tempnam() concern doesn't match anything in this PR's actual commit history or diff — looks like a stale reference to PR #591 rather than this PR.

LGTM.

debitor/ordreliste.php
```diff
@@ -56,8 +56,7 @@
 // 20260630 CDX/NTR Fixed land (country) column from printing the countries outside the table and searchable bar not existing.
 // 20260701 Sawaneh Fixed: 'Performed by' is display-only and no longer cleared on return to the list.
 // 20260701 CDX/NTR Fixed the default search to handle numeric comparisons and fixed TEXT searches from throwing fatal errors.
-// 20260911 CDX/LH SD-186 Label the searchable employee column Udført af in order and invoice lists.
-//                  Define it in the column pool so saved layouts use the same field configuration.
+// 20260911 LOE SD-685: filter selections are keyed, column setup follows the code.
 
 @session_start();
 $s_id = session_id();
```
ZaynSaul, line 62: Merge conflict here after master moved on (#630 added performed_by history lines — 20260916 CDX/LH and 20260918 CDX/PHR — at this same spot). Please rebase/merge master into this branch and keep both sets of history-comment lines (interleave chronologically), then push and re-request review.

ZaynSaul (changes requested):
After #630, #633, #639, #605 and #624 merged into `master`, this branch (still based on `a5280ddf`) now conflicts — GitHub shows `CONFLICTING` / `DIRTY`.

Checked with `git merge-tree`: only **one** file actually conflicts, `debitor/ordreliste.php`, and only in the file-history comment block at the top. Both this branch and `master` (via #630, "Preserve voucher input fields when opening document previews") added new history lines at the same spot, right after the existing `20260911 CDX/LH SD-186` line — this branch added the two SD-685 lines, master added two performed-by lines. No functional code conflicts: `grid.php`, `creditor_orderlist_grid.php`, `grid_account_lookup.php`, `grid_order.php`, the `lager/lister/*` files, and the new tests all merge cleanly.

Requesting changes purely to get the conflict resolved — not reopening the design/review discussion, which is already approved twice (mine + CodeRabbit). Please rebase/merge `master` in, keep both sets of history lines, then re-request review.

Loe4g:
@coderabbitai review

Loe4g:
@ZaynSaul Your conflict request is done, and CodeRabbit's findings are fixed as well — the description has the details.

* **Conflict:** merged `master` into `debitor/ordreliste.php`, keeping both history lines interleaved by date. It was the only file touched by `master`; nothing was rewritten.
* **CodeRabbit:** `8e5f9705` fixes legacy setups so omitted columns stay hidden, are normalised once, and remain visible/reversible in the editor. The account-lookup grid has no editor, so it keeps its existing append behaviour.
* **Follow-up:** `538e45bc` makes new setups current-format from the start (`visible => true`) and stops storing translated header text, so headers follow the session language.
* Since `538e45bc` is a new commit, CodeRabbit's approval of `8e5f9705` was dismissed; the current head needs a fresh review.

**Verified:** 52 tests / 108 assertions pass, the advisory script reports no mechanical findings, and live checks confirm new rows are current-format while legacy rows normalise exactly once.

Re-requesting your review.

ZaynSaul (approved):
Re-reviewed the current head (f39fa8bc) after the two review-fix commits and today's master merge.

Escaping (my finding from before, and the class of bug Lui flagged): confirmed fixed and didn't regress. `save_normalized_column_setup()`'s UPDATE escapes `tabel_id`, the JSON, and the compare-and-set `$stored` value, user id is cast to int - same as `save_column_setup`/`save_filter_setup`. Checked all three files that have these functions, not just grid.php.

Legacy/normalisation logic (`grid_setup_is_legacy`, `normalize_legacy_column_setup`): read through it directly. The compare-and-set UPDATE (`AND column_setup = '$stored'`) is real, not just described - a concurrent save from another tab loses the race cleanly instead of clobbering, and the setup just stays legacy for the next load. Existing rows are preserved verbatim; only missing code columns get added, hidden, with the `addedSinceSetup` marker.

The 538e45bc fix (new setups no longer written in the legacy shape): confirmed consistent across all four `fetch_grid_setup()` copies (grid.php, grid_order.php, grid_account_lookup.php, creditor_orderlist_grid.php) - all four now write `visible => true` and drop `headerName`/`description` on first insert. No partial-fix gap between the copies.

Output escaping on the rows the normalisation adds to the column editor: confirmed - field and align go through `htmlspecialchars(ENT_QUOTES, 'UTF-8')`, width is cast. 

Today's master merge (f39fa8bc): diffed it against 538e45bc directly - zero changes to any of the grid files, clean no-op merge as far as this PR's logic goes.

Two things I'd leave as follow-ups, not blockers:
- `grid_setup_is_legacy(array())` returns `false` for an empty array, so a legacy user who'd deleted every column (old editor could store `[]`) will see everything reappear once, instead of staying empty. Narrow, one-time-per-user edge case.
- A JSON-decode failure on a corrupted stored row now gets normalised and written back via `save_normalized_column_setup` (previously decode failures were read-only). Harmless in practice but worth a comment - a corrupted row silently becomes a synthetic one.
- Noticed in passing, not from this PR: the normal (non-hidden) column-editor row still echoes `customHeaderName`/`customDescription` unescaped - same as the old `headerName`/`description` did on master before this PR touched the file, so not a regression, but worth its own small follow-up given the hidden-row path right next to it now does escape.

Decision: Approve.
Blockers: none.
Testing: verified by me - escaping across all 3 save-path files, the compare-and-set UPDATE, the 4-way consistency of the first-write fix, the output-escaping lines, and the merge diff, all read directly against the current head. Not independently re-run by me: the PR's own live-tenant scenarios and the PHPUnit suites - taking those as author-reported.

PR #598, opened 2026-09-13
SST-769: save a currency rate without posting, and show the adjustmen…

momocoder14:
SST-769 adds an explicit “save rate without posting” action and makes DKK exchange-rate adjustments visible in account reports and CSV exports.

The rate-only action inserts or updates `valuta` without posting adjustment rows or recalculating account balances. `genberegn()` now runs only after a successful adjustment posting commits; it no longer runs after rate-only saves or rejected submissions. The existing posting action still creates the balanced adjustment. DKK adjustments are labelled in the on-screen account report and exported as numeric amounts in CSV.

Merged current master (`e9442a65`). External IDs are validated/cast, including binding an edited rate to its currency group. Translations use IDs 5153–5155.

## Validation

- `php tests/integration/currencyRateSave.php`: six actual POST-controller scenarios using an observable SQL double, with warnings enabled: insert/update without posting, ordinary first rate, ordinary adjustment posting, rejected date, and invalid difference account.
- The ordinary adjustment creates two posting rows, commits, and recalculates once. Rate-only and rejected cases never recalculate.
- The original head fails the rate-only regression because it recalculates once.
- PHP lint and whitespace checks passed. These new regression checks use a SQL double, not a live currency ledger.

## Reviewer scenarios

- `systemdata/valutakort.php`: insert and update rates without posting; compare `transaktioner` and `kontoplan` before/after. Then use ordinary posting and verify two balanced adjustment rows and updated balances. Reject invalid dates/accounts without changing balances.
- `finans/rapport_includes/kontokort.php`: confirm a nonzero adjustment displays a labelled DKK amount and the same numeric amount exports to CSV; verify an ordinary currency row and a zero amount.

The pre-existing rate-deletion path still calls `db_modify()` without its required context argument and fails under PHP 8. That separately identified bug, backdated adjustment rules, stale-adjustment recalculation, and voucher/journal traceability are outside this change.

## Summary by CodeRabbit

- **New Features**
  - Added a rate-only save option for currency rates without posting account adjustments.
  - Currency-rate updates now correctly save new and existing rate history entries.

- **Bug Fixes**
  - Currency creation and rate validation now work correctly in supported scenarios.
  - Rate-adjustment rows display the actual DKK amount instead of `0,00`.
  - CSV exports include the correct DKK value.
  - Account balances are recalculated only when an adjustment is posted.

PR #600, opened 2026-09-14
Sst 778

momocoder14:

A decision document for SST-778, at `doc/bilagspulje/PRIVATUDGIFT_DESIGN.md`. **No code** —
the ticket puts implementation out of scope until the accounting behaviour is agreed.

Two commits: the document, and a correction to it (see *An error I made and fixed* below).

### The finding that shaped it

Most of what this feature needs already exists.

`includes/docsIncludes/insertDoc.php`, included from `docPool.php:557-575`, already creates a
journal line from a pool document: it allocates the next `bilag` within the fiscal year
(`:118`), allocates `pos` (`:114`), inserts the `kassekladde` line with a placeholder
`d_type`/`k_type` of `F` and amount 0, reads back the new line's id (`:122`), updates its
fields from POST, and writes the `documents` row linking the file.

So "transfer a pool document to a journal" is not new. **A private expense adds which
accounts go on the two sides, whose payable it is, and the VAT treatment — in place of that
placeholder.**

Two further things fall out of existing mechanisms rather than needing design:

- **Draft versus posted.** `kladdeliste.bogfort = '-'` is an open, unposted journal
  (`finans/kassekladde.php:551`, `:726`); lines live in `kassekladde` until it is posted. A
  line written into an open journal *is* the draft, so "no automatic final posting by
  assumption" comes free, and the period-lock trigger on `transaktioner` then applies at the
  right moment.
- **Traceability and exactly-once.** `kassekladde` has a `dokument` column but
  `transaktioner` does **not**, so the durable link is the `documents` table —
  `source`/`source_id`, plus `user_id` and `timestamp`. That gives actor, timestamp and the
  exactly-once key with no new flag and no new column, and it survives posting because the
  `documents` row outlives the `kassekladde` row.

`kassekladde` also already carries every field the accounting decisions map onto: `d_type`/
`k_type`, `momsfri`, `ansat`/`medarb`, `kladde_id`. None of the recommendations need new
structure.

### What the document contains

| Section | |
|---|---|
| §1 | What exists today, with file:line references |
| §2 | The six decisions — whose expense, offset account, journal selection, VAT, permissions, draft vs posted — each with options and a recommendation |
| §3 | Recommended flow |
| §4 | Exactly-once, derived rather than flagged |
| §5 | Cancellation and failure behaviour |
| §6 | Worked examples, including the mixed-receipt case ruled out of a first implementation |
| §7 | Eight concrete Prodtest scenarios |
| §8 | Seven open questions for product/finance |
| §9 | Implementation scope for separate estimation |

Where I deliberately did not decide: **the VAT treatment of a reimbursed private outlay**
(§2.4). A private outlay may or may not carry deductible VAT depending on what was bought and
for whom. Both are expressible via `momsfri`, and the document says the answer is finance's
rather than defaulting to something plausible.

## An error I made and fixed

The first commit claimed "the pool never creates a journal line; it fills in one the user
opened". That is wrong, and it inverted the design — I had `insertDoc.php`'s work written up
as new implementation.

My first check grepped only *inside* `docPool.php`; the creation path lives one `include`
away. Caught by re-verifying the claim rather than trusting that grep, and corrected in the
second commit: §1.1 rewritten, §9's scope changed from "New" to mostly reuse, §3 now names the
path that does the work.

Flagging it because the rest of the document is the same kind of claim about what does and
does not already exist, verified the same way.

## How I verified this

There is no code to test. What can be checked is whether the document's factual claims hold,
and every `file:line` in it was derived against master `44523819` at the time of writing:

| Claim | Checked by |
|---|---|
| `insertDoc.php` creates a journal line from a pool document | reading `:89`-`:127`; included from `docPool.php:557-575` |
| A journal line is a draft until posted | `kladdeliste.bogfort = '-'` at `finans/kassekladde.php:551`, `:726` |
| `transaktioner` has no `dokument` column | column list on a live tenant |
| `documents` carries `source`, `source_id`, `user_id`, `timestamp` | column list on a live tenant |
| `kassekladde` has `d_type`/`k_type`, `momsfri`, `ansat`/`medarb` | column list on a live tenant |
| Picking an open journal matches existing behaviour | `ompost()` at `finans/kassekladde.php:4319` |

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive design documentation for transferring pool documents into draft journal lines for private expenses.
  - Documented journal and document-link behavior, employee and accounting rules, VAT handling, permissions, and draft-posting requirements.
  - Defined transfer workflow, failure handling, duplicate prevention using content hashes, transactional revalidation, and atomic journal/document writes.
  - Included examples, production test scenarios, open finance and product questions, assumptions, scope, and effort estimates.


ZaynSaul (changes requested):
Thorough write-up, and the citations check out against real code/commits. But there's a self-contradiction that needs fixing before this can guide implementation: §3 step 5, the Prodtest scenario 8, and Q7 all state the transferred document "stays in the pool... not moved or deleted." But §3's own recommended flow reuses insertDoc.php, whose existing attach path (:390 rename, :451 delete from pool_files) removes the document from the pool — which the doc itself already acknowledges in §4.2. An implementer following the doc as-is would build a "stays visible in the pool" UI that the reused code path makes impossible.

Also, §9.1 references "eight Prodtest scenarios in §7" but §7 lists eleven — please reconcile.

CodeRabbit's other three findings (content-hash race, legacy-row hash backfill, rename+DB non-atomicity) are worth addressing in the design too, though lower severity since nothing's implemented yet.

No code changes here, so no live testing needed — just want the pool-survival contradiction resolved (either the document really does get removed and Q7/§6.1/scenario 8 need updating, or the recommended flow needs to not reuse the deleting attach path) before this is used as an implementation reference.

LuiHedlund02:
@momocoder14 author action first: resolve Saul’s current-head pool-survival contradiction. The proposed reused insertDoc.php path removes the pool file/row; the document cannot simultaneously promise that it remains there without a different flow. Reconcile the scenario count and the hash-race/backfill/filesystem-vs-DB design questions, then request @ZaynSaul for re-review. Keep finance implementation choices explicit and pending domain sign-off.

LuiHedlund02:
@ZaynSaul @momocoder14 — pushed review fixes as `36a14326` (master merged into the branch first, clean).

**1. Pool-survival contradiction — resolved in favour of the code.** Verified on the merged branch: `includes/docsIncludes/insertDoc.php` renames the pulje file into `/finance/<kladde>/<source>/` (`:384`), inserts `documents` (`:398`), then deletes the `pool_files` row and the `.info` file (`:445-:451`). The document therefore *does* leave the pool on transfer. §3 step 5, the §5 failure table, the §6.1 worked example, Prodtest scenario 8, Q7 in §8 and the §9 scope table were updated to say so, and "transferred but still visible in the pool" is now stated explicitly as out of scope — it would need a transfer path that does not reuse `insertDoc.php`.

**2. Scenario count.** §9.1 said eight; §7 lists eleven. Fixed to eleven.

**3. CodeRabbit's three design-level findings** are answered in a new **§4.5 Open risks, and how they are mitigated**, grounded in what SST-740/#596 already shipped:
- *Content-hash race* — the hash check is a lookup, not a lock. Mitigation: a partial UNIQUE index on `documents (content_hash)` for journal sources, added idempotently in `betweenUpdates.php` the same way `pool_files_filename_uidx` is (`:571-621`), with the insert conflict as the decision point. The §4 explicit override writes its second row with a NULL hash plus a marker, so it cannot silently ride the same key.
- *Legacy rows with `content_hash = NULL`* — backfill in the same `betweenUpdates.php` pass via `hash_file()` over files still on disk, gated on a `settings` marker exactly as `pool_files_norm_amount_backfilled` is (`:69-88`). Rows whose file is gone stay NULL; that residue should be measured on a real tenant, not assumed small.
- *rename + DB non-atomicity* — reorder so the irreversible filesystem step comes last: line and `documents` row in one transaction, rename after commit, `pool_files` row last. A crash between commit and rename leaves a recoverable state that SST-740's sync pass with its 60-second grace window already tolerates (`docPool.php:200-216`). Prodtest scenario 11 covers the mid-flight case; per-step failure tests belong with the implementation.

§9.1 keeps the same total; the two assumptions that move it (Q7 answered against the move, legacy rows needing a controlled block) are listed explicitly.

ZaynSaul (dismissed):
Checked the rewritten §3/§4.5 against the actual code: rename in insertDoc.php:384, documents insert ~:397, pool_files/.info cleanup :445-451, and the pool_files_filename_uidx pattern in betweenUpdates.php:595-617 all match what's cited. The doc now correctly states the transfer removes the document from the pool rather than leaving it, and §4.5's proposed reorder (documents insert inside a transaction, filesystem rename after commit) is a sound fix for the atomicity gap. Scenario count and Q7 wording are consistent with the rest of the doc. LGTM for a design doc.

doc/bilagspulje/PRIVATUDGIFT_DESIGN.md
```diff
@@ -0,0 +1,515 @@
...
+
+| Piece | Rough shape |
+|---|---|
+| Checkbox and dialog in the pool UI | `includes/docsIncludes/docPool.php`, and `finans/pulje_review.php` if the mobile screen gets it too |
+| Draft line creation | **Mostly reuse.** `insertDoc.php` already creates the line, allocates the voucher and writes the `documents` row (§1.1). The work is setting the two account sides, the employee and the VAT treatment instead of the `F`/`F`/0 placeholder |
+| Payable account setting | `settings` table; no migration if an existing group is reused |
+| Exactly-once check (§4) | A `pool_transfers` table with `UNIQUE (content_hash, forsoeg)` and its idempotent migration in `includes/betweenUpdates.php`, hashing at transfer time, and the insert conflict handled as the decision point rather than a `SELECT` (§4.5) |
+| Transferred-but-visible pool entry | Out of scope (§3 step 5). Only in play if Q7 is answered against the move behaviour, and then it is a new transfer path, not a marker |
+| `findtekst` ids for the dialog | Coordinate with open branches — PR #447 (`feature/udfoert-af`) holds 5151–5152, SST-769 holds 5153–5155 |
```
ZaynSaul, line 464: Minor staleness, non-blocking: PR #447 is closed (holds nothing anymore), and SST-769 (#598) moved its 5155 to 5234 on 9873a827 — about 14 minutes after this commit. Worth a one-line update before merge.

ZaynSaul (approved):
Re-reviewing from scratch since my prior LGTM got auto-dismissed by the push after it (`87ff272f`, "the exactly-once key cannot live on documents").

That commit itself corrects a real problem in Lui's `36a14326` mitigation, and I verified it against the actual code rather than taking the commit message on trust:

- **`documents` holds several rows per file by design — confirmed.** `insertDoc.php:398` inserts the first `documents` row, and the sibling-attach paths at `:419` (explicit `targetSourceIds`) and `:434` (bilag-fallback) insert additional rows for the *same* filename/hash under `source = 'kassekladde'`. A UNIQUE constraint on `documents(content_hash)`, even restricted to journal sources, would reject the second sibling insert and break that shipped multi-attach behaviour. The doc's line numbers are off by a line or two from current master (`:420`/`:432` vs actual `:419`/`:434`) — trivial drift, not a citation error.
- **Restricted/partial unique index is Postgres-only — correct.** MySQL/MariaDB `CREATE INDEX` has no `WHERE` predicate, which is exactly why `betweenUpdates.php:589-595` branches on `$db_type` for the existing `pool_files_filename_uidx` pattern the doc cites. Confirmed that branch exists as described.
- **`pool_transfers(content_hash, forsoeg, ...)` with `UNIQUE (content_hash, forsoeg)`** is the right shape for the fix: it's per-transfer rather than per-attachment, portable to both engines, and the `forsoeg=2` override path keeps the explicit-duplicate case queryable instead of hiding it behind a NULL hash (which was the weaker part of the previous revision). Sound.
- Section renumbering (nine questions, Q9 added, §4.5 rewritten, §9.1/§9.2 estimate rows updated) is internally consistent on this read.

**One small, non-blocking staleness to fix in the same spirit as the doc's own "verify, don't trust" standard:** the §9.1 coordination note still says "PR #447 (`feature/udfoert-af`) holds 5151–5152, SST-769 holds 5153–5155." Checked both: #447 is closed (not merged, so it holds nothing), and SST-769 (#598) renumbered its 5155 to 5234 on `9873a827` — which landed about 14 minutes after this commit, so it wasn't stale when written, just overtaken since. Worth a one-line update before merge, but it's a footnote in a future-estimate table, not something blocking agreement on the design.

No code here, so nothing to test live. Approving — the accounting design, the exactly-once mechanism, and the citations all hold up against current master.

PR #601, opened 2026-09-14
SST-774: design document for backdated currency postings and corrections

momocoder14:
Decision document only - no implementation, per the ticket's scope. Placed in doc/valuta/ following the doc/stripe/INTERFACE_CONTRACT.md precedent.

The finding that shapes the design: Saldi already has the audit-preserving correction mechanism this ticket asks for. ompost() in finans/kassekladde.php copies a posting into an open kassekladde with debet and kredit swapped, and the user reviews and posts it; the original is never touched. It cannot be used on rate adjustments because those are written straight into transaktioner with kladde_id='0' and never exist in a kassekladde.

So traceability is not a separable nice-to-have. Without a kladde_id the existing correction flow cannot be pointed at these rows at all, which is why the recommendation is to create the adjustment as a kassekladde entry rather than two direct transaktioner inserts. That one change yields a real bilag number, makes ompost() applicable, gives review-before-post, and handles closed periods correctly - the moms_periode_luk trigger refuses at posting time with the entry still in the journal.

Also documented: staleness derived rather than stored; the boundary and cascade cases (entry dated on the adjustment date, a later rate change made stale by correcting an earlier one); why automatic recalculation is ruled out by the customer and by the ticket; and that transaktioner has no row-to-row reference column, so linking a correction to its original needs a column added idempotently in includes/betweenUpdates.php rather than a convention in free text.

Seven questions for product/finance in section 6. The document is blocked on those answers; implementation is listed in section 7 for separate estimation.

Line references verified against master 44523819.


ZaynSaul (approved):
Solid design work — verified all cited line references against the stated baseline commit, still accurate against current master. The correction mechanism (posting adjustments via kassekladde and reusing the existing ompost() reversal rather than in-place edits) is audit-preserving and reuses a proven pattern. The cascade case (§4.3) and closed-period handling are both correctly addressed.

Consistent with #598, which appears to already implement the two immediate fixes this doc describes as delivered.

Only gap: multi-currency rounding is mentioned incidentally but not discussed as its own design question — minor, and the doc is already honest that it's blocked on 7 open questions pending product/finance sign-off. Approving as a design doc; implementation should wait for those sign-offs as stated.

PR #604, opened 2026-09-14
SST-784: Escape form text only at the PostScript output boundary

LuiHedlund02:
[SST-784](https://saldi-team.atlassian.net/browse/SST-784): Shared `skriv()` escapes parentheses before emitting both PostScript and HTML, so HTML/PDF shows literal backslashes. Escape parentheses and existing backslashes only at the PostScript output boundary; keep the HTML text unchanged.

This implements the confirmed escaping defect. The ticket’s premature-wrapping report remains open pending confirmation of manual line breaks and template width.

## Validation
- `php tests/integration/printTextEscaping.php`: five text fixtures across left/right/center alignment, checking actual HTML output and PDFs rendered with Ghostscript. Includes nested/unbalanced parentheses, literal backslashes and ordinary text. All 15 cases pass with warnings enabled.
- The same test fails against current master because HTML contains escaped parentheses.
- PHP syntax and whitespace checks pass.

## Docker application verification — 2026-09-14

Six full-application PDF cases passed with PostgreSQL and real converters: quote, order confirmation and invoice, each in HTML and PostScript mode. All preserve literal parentheses and Windows-style backslashes, with the expected page/line count. The baseline HTML confirmation reproduces the unwanted escaped parentheses. The 15 integration cases also pass. The separate wrapping question and customer-template validation remain open.

## Remaining review and customer verification
- `includes/formfunk.php::skriv`, called by order printing and shared form text: print an order confirmation containing parentheses, a Windows-style path and nested parentheses with HTML generation on and off; compare visible text, alignment and page breaks.
- Exercise a long description crossing a page boundary and bold/italic text in quotes, invoices and confirmations; verify headers/footers and totals remain unchanged.
- Ask the customer whether the wrapping example contains manually entered line breaks, then inspect the affected template width. This PR does not change wrapping.

Assignee: Momodou Jammeh. Customer/Prodtest verification remains; no production deployment.

- [ ] I have updated relevant wiki documentation (not needed for this fix).
- [x] No new external styles or scripts.
- [ ] I have read the externally linked developer guidelines.

Repository doc/ai/MEMORY.md and linked conventions followed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved PostScript text rendering for parentheses and backslashes.
  - Preserved original text in HTML output.

- **Tests**
  - Added integration coverage for special characters across supported text alignments.
  - Verified that rendered HTML and converted PDF output match the original input text.


ZaynSaul (dismissed):
The fix is correct — single-pass strtr() escaping avoids double-escaping backslashes, and all three PS-special characters (\, (, )) are covered. Well-validated already (unit tests + Docker/Ghostscript rendering).

One gap worth a follow-up: skriv() has a second PS-emission path for the page-break "formular variabler" text (formfunk.php:328, built from $row['beskrivelse'] template tokens substituted with DB fields) that was never covered by the old escaping and still isn't covered by the new strtr call. If a substituted field (company name, address) contains a parenthesis, it'll hit the same defect class this PR is meant to close.

Approving the fix as scoped — please open a follow-up for the formfunk.php:328 path, and live-test printing an order/invoice whose template triggers that page-break path with a parenthesis in a substituted field.

LuiHedlund02:
@ZaynSaul rebased onto master; the conflict was header history only. The real HTML/PostScript regression passes all five text fixtures in three alignments through Ghostscript. Please re-approve the new head. Review focus: escaping stays at the PostScript output boundary. Your page-break template-variable escaping finding remains a follow-up; test a page break with parentheses in substituted company/address text.

ZaynSaul (dismissed):
Re-reviewed the rebased head. The core PostScript-escaping fix is byte-identical to what I already approved - the rebase conflict really was header-history only, as you said.

The new integration test is a genuine improvement - I ran it directly (not just read it): all 5 text fixtures x 3 alignments pass through real Ghostscript rendering. On CodeRabbit's two new findings against the test file, I checked both empirically and they're false positives:
- "Duplicate utf8_iso8859() declaration" - doesn't actually conflict. PHP hoists the test file's unconditional top-level function declaration at compile time, before the require_once(formfunk.php) executes, so by the time formfunk.php's `if (!function_exists('utf8_iso8859'))` guard runs, function_exists() is already true and it skips its own definition. No redeclaration error.
- "Uninitialized skriv() globals under the strict error handler" - also didn't reproduce; the test ran clean with no ErrorException across all 15 calls.

My original formfunk.php:328 page-break ($ny_streng) gap is still open as you noted - agreed that's a fair follow-up, not a blocker here.

Approved.

LuiHedlund02:
Rebased onto master after #605 merged (header-history conflict only) and folded in the follow-up you flagged: the page-break "formular variabler" PostScript write in skriv() now goes through the same strtr() escaping as the main path. Ran tests/integration/printTextEscaping.php in the saldi-web container: all 5 fixtures x 3 alignments pass through real Ghostscript. Re-requesting since the push dismissed your approval.

ZaynSaul (approved):
Verified against the actual branch head (not just this incremental diff): the old blanket `str_replace("(", ...)` at the top of `skriv()` is gone, and both PostScript emission sites — the main line-text branch (already fixed by 2a7c3ae7) and this PR's page-break/"formular variabler" branch — now escape backslash-then-parens only at the PS `fwrite` boundary, leaving HTML output untouched. No other unescaped `moveto (...) show` site remains in the file. LGTM.

PR #579, opened 2026-09-09
SST-780: Preserve document filenames when merging PDFs

LuiHedlund02:
## What are the changes about?
SST-780: Multi-page quotes and order confirmations currently replace their document filename with `udskrift.pdf`. Preserve the original name in both HTML and PostScript merge paths. Merge through a temporary PDF before replacing a destination that may also be an input, and clean source pages only after success. Conversion/merge failures stop processing instead of publishing a partial merge.

https://saldi-team.atlassian.net/browse/SST-780

## Local verification — passed, 9 September 2026

Verified against an isolated, disposable MEDSHOP dataset clone using PHP 8.3, PostgreSQL 16, WeasyPrint 62.3, Ghostscript 10.05.1, real pdftk and Playwright.

- Eight full browser flows through `debitor/formularprint.php` → `includes/formfunk.php` → `includes/udskriv.php`: quotes and order confirmations, short/long orders, HTML printing on/off. Each yielded the expected document filename and a valid PDF over HTTP; short documents had one page and long documents three. A rendered PDF page was visually inspected.
- Before/after comparison reproduced `udskrift.pdf` on baseline and the correct numbered document filename on this branch.
- Both actual conversion branches additionally passed real-renderer one-/three-page fixture checks, verifying rendered text and that the final PDF survives cleanup.
- Integration checks passed for page order, filenames with spaces, destination also being an input, missing/corrupt input preserving the original and removing temporary output. Run `php tests/integration/mergePrintPdfs.php` with pdftk and ps2pdf installed.
- PHP syntax and whitespace checks passed.

The clone lacked its logo file; a neutral EPS fixture was supplied for local printing. Customer branding, physical printing and production deployment are not covered by this verification. Customer data/PDFs/screenshots remain local and are not attached to this public PR.

Regression surface: quote/confirmation generation, shared PDF conversion/merging, temporary-file cleanup and HTTP document links. Other document types using the shared merger should retain their original names too; invoice/statement business workflows were not exercised.

## Have you checked the following?
- [ ] I have updated relevant documentation at https://github.com/DANOSOFT/saldi/wiki
- [x] I have checked that I am not linking to new external styles or scripts.
- [ ] I have read and understood the externally linked developer guidelines.

Repository `doc/ai/MEMORY.md` and linked conventions were followed. No wiki change is needed for this bug fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF generation and merging reliability, including safer handling of file paths and conversion failures.
[...]

PR #605, opened 2026-09-14
SST-789: Print all selected invoices in one ordered PDF

LuiHedlund02:
[SST-789](https://saldi-team.atlassian.net/browse/SST-789): Printing several selected invoices can return only the last invoice in the PostScript path. Preserve the ordered generated-document list and page counts in the session, scoped to the current tenant/user/output. Render all selected invoices into one atomically published `faktura-batch.pdf`.

HTML pages follow the generated numeric page order. The PostScript source already contains all pages, so its HTML counterparts are not appended again. Each invoice retains its own PDF letterhead. Missing pages, conversion failures and merge failures stop the request with an error and return link; existing output and sources are preserved. Single-document and email branches retain their existing routing.

## Validation
- Real Ghostscript, pdftk and WeasyPrint integration checks: one/two/four invoices, including a ten-page invoice, different selection versus filename order, filenames with spaces, exact page counts and content once per invoice/page.
- Per-invoice letterhead; missing primary/extra page, converter failure, merge failure, path traversal, missing letterhead and corrupt PostScript. Failures preserve output/sources and clean temporary workspaces.
- Actual generator output and converter gate characterized for one/four invoices, return-query preservation and tenant/user/document mismatch rejection.
- PHP syntax and whitespace checks pass.
- Commands: `php tests/integration/printBatchHandoff.php`; `SALDI_TEST_REAL_WEASYPRINT=1 php tests/integration/printBatch.php`. Without the environment flag, HTML conversion uses a deterministic fixture adapter while PDF conversion/merging remains real.

## Docker application verification — 2026-09-14

Real PostgreSQL/application/converter checks pass for two- and four-invoice batches and reordered selections in both HTML and PostScript modes. A 70-line invoice occupies six pages; the four-invoice batch has nine pages, every selected line exactly once, correct order and backgrounds on every page. The actual invoice-list print POST and redirect also produce the nine-page PDF. Selection fields were supplied by the harness because the copied account's datagrid configuration omitted the checkbox column. Baseline e852d8c reproduces the PostScript bug: only the final invoice is returned.

Short single-invoice checks pass in both modes; the long single invoice passes in HTML. **Existing failure:** printing the long invoice alone in PostScript yields 11 pages instead of six, duplicating continuation lines. Baseline and this PR give identical results. This is the unchanged legacy single-invoice path and still needs a separate fix. The batch renderer's 14 integration scenarios and handoff scope checks pass with real converters. No email was sent; custom production converters and customer configurations remain unverified.

## Remaining review and customer verification
- Invoice list → `debitor/formularprint.php` → `includes/formfunk.php`: select one, two and four invoices, including a long invoice. Verify resulting PDF page count, each invoice once and selection order with HTML generation on/off.
- `includes/udskriv.php` / `renderPrintBatch`: test separate department/language letterheads, PDF-tekst, file download and configured custom converter. Force a conversion failure and confirm no partial PDF link is served.
- Repeat a single-invoice print and a batch email operation, and check return navigation and a second user/session. The Docker application/browser results are recorded above; email, custom-converter and customer verification remain pending.

Related PR #579 (SST-780) touches the same conversion dispatch in `includes/udskriv.php`; reconcile that overlap when merging. This PR does not include its commits. Assignee: Momodou Jammeh. No production deployment.

- [ ] I have updated relevant wiki documentation (not needed for this fix).
- [x] No new external styles or scripts.
- [ ] I have read the externally linked developer guidelines.

Repository doc/ai/MEMORY.md and linked conventions followed.


ZaynSaul (dismissed):
Nice work tracing through this — the session-scoping check (udskriv.php:69-73), page-count logic, and per-invoice letterhead isolation all trace out correctly against the old code, and the letterhead handling here is actually safer than the pre-existing mail-branch code (which reuses one background PDF across all mailed invoices — pre-existing, out of scope, but flagging in case it's worth its own ticket). No command-injection/path-traversal issues in renderPrintBatch.php — all paths are basename()/realpath()-checked and shell-escaped.

Two nits, non-blocking:
1. The new error message ('PDF-udskriften kunne ikke oprettes...') is missing the ID| prefix other findtekst() calls use, so it won't be translatable.
2. #604 also touches formfunk.php near the same anchor lines — expect a trivial rebase conflict, not a functional issue.

Approving — please still generate a real multi-invoice PDF (mixed HTML/PS, mixed letterheads, an odd-page-count invoice) through the app before merging, since page-bleed/ordering issues are hard to catch statically.

LuiHedlund02:
@ZaynSaul rebased after #604 and initialized `pages` on every appended batch document. The converter still gets the completed page count after rendering. Validation: 14 real-converter scenarios and both single/multiple-document session-handoff scenarios pass. Please re-review the new head. Merge order: #604 first, then this PR. Manual release check remains mixed HTML/PS, mixed letterheads and an odd-page-count invoice; verify order, no page bleed and stale session isolation.

ZaynSaul (approved):
Re-reviewed the rebased head plus the new fix. Traced it through the code: $printBatchDocuments[] entries were being appended with no 'pages' key at all, so renderPrintBatch()'s HTML multi-page loop (for ($page = 2; $page <= $document['pages']; $page++)) would read an undefined key as null and never iterate past page 1 for any document whose count didn't get backfilled before use. Defaulting to 'pages' => 1 at creation, with the existing max(1,(int)$side-1) update still applying afterward, closes that correctly.

One caveat: I couldn't independently run the converter integration tests here - this sandbox doesn't have pdftk or weasyprint installed and I don't have a way to install them. So this is verified by code-tracing, not by reproducing your 14 real-converter scenarios myself. Flagging that gap rather than claiming full re-verification.

Approved.

PR #607, opened 2026-09-14
MB-48: Fix cash checkout hanging when automatic receipts are disabled

LuiHedlund02:

Fixes [MB-48](https://saldi-team.atlassian.net/browse/MB-48): with automatic receipts disabled, a walk-in cash sale was booked but remained on the change screen. Reproduced locally with a 25 DKK sale and 100 DKK received: the screen showed 75 DKK change, while the order already had status 3 and one cash payment.

`afslut()` checked an undefined `$printpopup` variable, skipped its only drawer redirect, and then exited unconditionally. Remove that condition so checkout sends the drawer-only request (`skuffe=1`, empty receipt) with the return URL for the next sale. Also avoid mutating `SERVER_NAME` while building that URL and handle missing HTTPS/cookie values without warnings.

Remote drawer endpoints now require HTTPS. Bare remote hostnames/IPs default to HTTPS; explicit remote HTTP URLs and malformed origins are rejected with a visible configuration error and a link to the next sale. HTTP remains available for explicit loopback endpoints (`localhost`, IPv4 loopback and `::1`), and Android keeps its `saldiprint://` handler. Redirect parameters are URL-encoded and the HTML attribute is escaped. There is no HTTP retry if a remote printer does not support TLS.

## Validation

- `php tests/test_pos_cash_checkout.php`: **37 scenarios pass with E_ALL and no warnings**, run using the local `saldi-web` PHP image with networking disabled. Covers cash/change, exact cash, split card/cash, card/change, exact card, account customers, automatic receipts, Android, printer fallbacks, remote HTTPS, loopback HTTP/HTTPS, ports, IPv6, and rejected remote HTTP/malformed/lookalike origins. All scenarios verify settlement and commit before navigation or the configuration error. The expanded suite fails 29 scenarios against the previous PR head, confirming regression coverage.
- PHP lint and `git diff --check` pass.
- Prior browser tests, before HTTPS enforcement, against an isolated local Postgres tenant confirmed that Kontant with change, exact cash, and Kontant på beløb return to a fresh checkout. Each completed order has exactly one payment. The existing on-amount path records the exact balance, even when a higher amount was entered.
- Browser testing used a simulated print server; physical drawer/printer operation and TLS support on the installed remote print server remain to be verified. The simulator mapped the callback to local port 5000 because the legacy callback builder omits the port. The customer's backup was not available; reproduction used a local tenant with the relevant receipt setting.

The standalone regression test executes the real `afslut()` in child processes with simulated database, voucher, settlement and receipt boundaries. It checks commit before navigation; it does not validate accounting internals or physical hardware.

## What to test before release

- **POS checkout / `afslut()`:** automatic receipts off, sell for 25 DKK and receive 100 via Kontant. Verify 75 change, drawer opening, a fresh checkout, and exactly one payment. Repeat with exact cash and Kontant på beløb.
- **Other completion paths:** try split card/cash, exact card and an account customer; verify their existing navigation and drawer behavior.
- **`pos_txt_print()` / printer integration:** enable automatic receipts and verify receipt output and return to checkout. Check the installed receipt printer or Android handler on the relevant till.
- **Drawer transport:** with automatic receipts off, verify local HTTP (`localhost`/loopback) and remote HTTPS drawer opening and return to checkout. A remote printer must support TLS with a certificate trusted by the till browser. An explicit remote `http://` setting must show the configuration error, send no drawer request, and allow continuing to the next sale without booking the completed sale again.

- [ ] I have updated relevant documentation at https://github.com/DANOSOFT/saldi/wiki — not applicable; existing checkout behavior is restored.
- [x] I have checked that I am not linking to external style or JavaScript files that could compromise stability.
- [ ] I have read and understood the [developer guidelines](https://docs.google.com/document/d/1GOmomtvKf21OV2VWNOIPweDi4gJHpMCK5qXzrPrypUc/edit?usp=sharing). Repository `AGENTS.md` and linked project conventions were reviewed; the separately linked document was not reviewed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed cash checkout navigation so the transaction is committed before redirecting.
  - Improved redirect handling for both HTTP and HTTPS checkout sessions.
  - Ensured cash payments without automatic receipt printing continue to the appropriate print server or return page.
  - Improved cash drawer opening and receipt-printer fallback behavior.


ZaynSaul (approved):
Traced the root cause precisely: $printpopup is checked in exit.php's drawer-redirect gate but is never set anywhere in this code path (it's an unrelated same-named local in other files) — so the redirect never printed and the function fell straight to exit;, hanging the checkout screen. The unconditional-print fix directly addresses that; it's not a workaround/timeout.

Bonus find: the old $url="://".$_SERVER['SERVER_NAME'].=$_SERVER['PHP_SELF']; had an operator-precedence bug (.= instead of .) that silently corrupted $_SERVER['SERVER_NAME'] for the rest of the request. Fixed correctly.

The new cashDrawerPrintOrigin() HTTPS-enforcement is well-built — rejects userinfo/SSRF-style bypasses, non-http(s) schemes, and case-mismatched loopback hosts, with good test coverage (37 scenarios).

Approving on the code — please still do a real click-through on an actual till before this ships, per your own PR notes: cash sale with auto-receipts OFF (drawer opens, clean return to checkout) and ON (receipt prints, unaffected), since hardware/TLS on the real print server wasn't tested, only simulated.

PR #609, opened 2026-09-14
SST-790: a zero price leaves the label field empty

Loe4g:
dkdecimal() has no blank-for-zero branch, so an item priced 0,00 printed "0,00" on price labels. 9f756f03 covered $minpris, but the reported layout uses $pris, which was still substituted unconditionally - and the mit-salg branch compared the formatted "0,00" with 0, which is false under PHP 8.

- $pris and the $minpris fallback now share one zero-aware value ($vispris) and print an empty field for 0,00; same rule for the legacy oldlabel.php.
-- $dkkpris follows the same zero rule in both renderers. It was substituted unconditionally, so a layout using it printed "0,-" for a 0,00 item even after the other price fields went blank (raised in review). A new addition from Review, couldn't be verified from UI, please see below.
- The mit-salg branch tests the raw mylabel price (kept in $rawprice per cell) instead of the formatted value.
- NewlabelCharacterizationTest: pins a plain item-card print filling the template's grid (the sheet 9f756f03 restored) and an empty price field for $pris and $minpris in both branches, with a non-zero control.

Verified against test_11 on ssl12 (item 2524, salgspris 0.000, layout "Pris $pris") and test_11(both layouts): zero prints an empty field, real prices unchanged.

##PS
The "_______" placeholder has been removed so the result is consistent with the sample provided by the customer. It also avoids the placeholder being visually mistaken for part of the barcode, as it was positioned too close to it.
## Scope note

$dkkpris is not used by any layout we can see: 0 of 11 stored layouts across the local databases, no importfiler/ template, no settings/tekster row, and the visual editor's generator only emits $minbeskrivelse/$minpris. It is fixed here because it is the same defect class; the only visible change for a tenant that does hand-write it is a blank price field at 0,00 instead of "0,-".

## Summary by CodeRabbit

- **Bug Fixes**
  - Zero-priced items now leave price fields blank instead of displaying “0,00” or placeholder characters.
  - Price formatting remains unchanged for items with non-zero prices.
  - Plain and legacy label prints now correctly render expected price fields and template grid cells.

- **Tests**
  - Added coverage for zero-priced items, formatted prices, full-sheet printing, and single-cell label printing across both label renderers.


Loe4g:
@coderabbitai review

Loe4g:
@coderabbitai resolve

ZaynSaul (changes requested):
The logic fix is correct — the old truthy-check on the post-formatted $price string (e.g. "0.00") failed to detect zero under PHP 8's string-comparison rules; switching to $rawprice with is_numeric(...) && == 0 fixes it, and the same pattern is applied correctly to the $vispris substitution. CodeRabbit's two findings are also genuinely resolved on the current head.

However, GitHub reports this PR as conflicting with master (mergeable: CONFLICTING). #614 (already merged) rewrote the exact same block in lager/labelprint_includes/newlabel.php, adding a third branch ahead of the ones this PR touches. A naive merge risks losing #614's commission-blank behavior or reintroducing the zero-price bug in that path.

Please rebase onto current master and manually reconcile the branch ordering between #614's commission-blank logic and this PR's $vispris/$rawprice changes, then live-verify both behaviors coexist (a commission-blank label and a zero-price label both print correctly), plus a normal non-zero price through both newlabel.php and oldlabel.php.

LuiHedlund02:
@Loe4g Saul’s current-head review is blocked on conflicts after #614 changed the same label code. Rebase preserving the latest layout, then request @ZaynSaul for re-review. Verify every displayed price variant, including $dkkpris, and explain the placeholder removal so the final description matches the behavior.

Loe4g:
@LuiHedlund02 and @ZaynSaul this has been updated. Please read the description for more context.

Loe4g:
@coderabbitai ressolved

ZaynSaul (commented):
Following up on Lui's ask for a re-review after the rebase past #614.

**Rebase — done, verified.** `143fe6c7`'s merge-base with master (`8f9cd14f`) descends from #614's merge commit (`805aaa4f`) - confirmed via `git merge-base --is-ancestor`, not just the mergeable flag.

**Every displayed price variant, including `$dkkpris` — done.** `$pris`, `$minpris` (both the mit-salg/commission branch and the plain branch), and now `$dkkpris` all blank correctly on a zero price, in both `newlabel.php` and `oldlabel.php`. I also checked `$special_pris` (`$r['special_price']`) as a possible missed variant - it's not one: that field is never run through `dkdecimal()`, so it doesn't have the "0 formats to the string '0,00'" defect this ticket is about. Correctly left untouched.

**Tests — ran them, not just read the diff.** All 15 cases in `NewlabelCharacterizationTest.test.php` pass (29 assertions): zero/non-zero controls for `$pris`/`$minpris`/`$dkkpris` across both renderers and both mit-salg branches.

**Still outstanding: explain the placeholder removal.** The PR description doesn't yet explain why `$minpris = "_______"` was removed. I traced it: that underscore wasn't a deliberate "write the price in by hand" affordance - it's itself a prior, cruder fix for this exact same 0,00-display defect (commit `9f756f03`, "Fixed labels, 0,00 prints...", by sirrolin). So removing it in favor of a true blank isn't a regression, it's superseding an earlier workaround with the correct fix - but as written, a reviewer skimming the description has no way to know that and could reasonably read it as an unexplained UX change. @Loe4g could you add a line to the description along those lines? Once that's in I'm ready to approve - the code and tests are solid.

ZaynSaul (approved):
Approving. Rebase past #614 is confirmed (merge-base ancestry checked), $dkkpris and every other displayed price variant are handled correctly, and all 15 characterization tests pass (verified by actually running them). The placeholder-history note I asked for is a nice-to-have, not a blocker - not holding up approval on it.

Loe4g:
Nice one @ZaynSaul I removed it because I initially mistook the placeholder for part of the barcode, and I thought other customers might too since the sample of the ticket was completely blank too. It was also too close to the barcode, so I wanted to avoid any potential scanning interference.

PR #618, opened 2026-09-15
Add session-based MySale customer entry and logout

PeterRude:
An already logged-in Saldi user can open `mysale/mysale.php`, enter a customer number and open that customer's existing MySale page. The shop comes exclusively from the current Saldi session in the master `online` table; URL shop/database parameters cannot choose the tenant. Missing, expired and customer-portal-only sessions go to the actual login form, `index/index.php`.

Add a CSRF-protected “Log af” button that clears customer authentication and filters, removes only temporary customer online rows, and returns to customer-number entry while retaining the staff session. MySale responses use `Cache-Control: no-store`. Existing customer-ID links and customer password checks remain in place.

Deploy both `mysale/mysale.php` and the new `mysale/shopLookup.php` together.

## Validation
- `php tests/test_mysale_shop_lookup.php`: 23 passing checks for session validation, tenant selection, closed accounts, customer lookup, encoded links, escaping and customer logout.
- PHP syntax checks for both application files and `git diff --check` pass.
- HTTP test confirms unauthenticated requests follow a 303 redirect to `index/index.php` and display the Saldi login form. Earlier HTTP checks also confirmed invalid session cookies cannot expose customer-number entry.

Manual checks before deployment:
- Log into a shop, open MySale without `id`, enter an existing customer number and verify the correct customer and existing password behavior.
- Click “Log af”; verify customer-number entry returns without another staff login, and the next customer does not inherit the previous customer's login or date filters.
- Verify unknown customer numbers, existing customer-ID links, and existing password-reset links.

The full authenticated browser/customer/logout round trip has not been tested end to end.

- Documentation: deployment requirements and manual scenarios are documented above.
- [x] No external stylesheets, JavaScript or other external resources added.
- [ ] External developer guidelines reviewed.


ZaynSaul (changes requested):
The security redesign is a real improvement — deriving the tenant strictly from the authenticated staff session instead of trusting request params. Escaping, CSRF (hash_equals), XSS (htmlspecialchars), and PRG pattern all check out clean.

One likely regression: mySaleShopLookup() picks the session's online row via ORDER BY logtime DESC LIMIT 1 and rejects it if rettigheder is empty. But the existing id-flow inserts a second online row under the same session_id with rettigheder='0' when a customer is opened via ?id=. If staff later reopens mysale.php without id/tmpcode (bookmark, back button, without clicking "Log af"), the query now returns the customer row instead of the staff row, and staff gets wrongly bounced to the login page despite a valid session.

Suggest filtering the query to rettigheder <> '0' (matches the existing convention around mysale.php:164), or keying off a dedicated staff marker instead of recency.

Please live-test: open a customer via id, then reopen mysale.php without logging off first, and confirm staff isn't kicked out. That's the exact round trip the PR description already flags as untested.

(Noting separately, not a regression: raw id links remain unsigned/unbound to session, per the description's own scope note — worth a follow-up ticket.)

LuiHedlund02:
@PeterRude Saul’s current-head change request needs author work before Nicolai’s security/lifecycle review: distinguish the staff online row from the newer customer row (`rettigheder=0`). Live-test opening via id and reopening mysale.php without logging off; staff must stay authenticated. Preserve Saul’s blocking review and request his re-review after the fix, then @SirRolin for the tenant/session and 24-hour entry semantics.

LuiHedlund02:
@ZaynSaul @PeterRude Review fix pushed as f32ebfe6.

**Problem (confirmed):** the id-flow in `mysale.php` inserts a second `online` row under the same `session_id` with `rettigheder='0'` when a customer is opened via `?id=`. `mySaleShopLookup()` picked the newest row for the session, so reopening `mysale.php` without `id`/`tmpcode` (bookmark, back button, no "Log af") returned the customer row and bounced staff to the login page.

**Change:** the session lookup in `mysale/shopLookup.php` now filters `AND rettigheder <> '0'` (same convention as `mysale.php:210`). The rest of the validation (db present, rights non-empty, 24h freshness) is unchanged.

**Tests:** `tests/test_mysale_shop_lookup.php` gained two assertions — the lookup SQL contains the `rettigheder <> '0'` filter, and a rights-0 row with a newer `logtime` does not shadow the staff row (the db mock now applies the filter, so the case genuinely fails without the fix).

Evidence (no host PHP; run in the `saldi-web` container):
- `php -l mysale/shopLookup.php` → `No syntax errors detected`
- `php -l tests/test_mysale_shop_lookup.php` → `No syntax errors detected`
- `php tests/test_mysale_shop_lookup.php` → 25 PASS, exit 0
- Negative control: with the filter removed the suite fails with `RuntimeException: Session lookup filters out customer rows in SQL`

No conflict with master (`mergeable: MERGEABLE`), so no merge commit was added.

ZaynSaul (approved):
Verified. `AND rettigheder <> '0'` in shopLookup.php's session query is the same filter already used in mysale.php:164, not a new pattern — so the staff-session resolution now matches the established convention and a newer customer row can no longer shadow it. New test covers the shadow case directly. LGTM.

ZaynSaul (changes requested):
Clarifying/correcting my prior approval: it covered exactly one thing — the shadow-row regression fix I flagged on 09-16 (shopLookup.php filtering rettigheder <> '0'). It was never meant to, and should not be read as, clearing the two separately-requested reviews Lui laid out in the same thread: Nicolai's security/lifecycle review, and SirRolin's tenant/session + 24-hour entry semantics review. Neither has happened yet — this PR currently shows reviewDecision APPROVED off my review alone, which overstates how merge-ready it is. Switching to Request changes to reflect that accurately.

**Decision:** Request changes (my part is done; two other named reviews are still outstanding and required before merge per Lui's 09-16 sequencing).

**Outstanding, not covered by my approval:**
- @Nicolai — security/lifecycle review of the session-based tenant-derivation redesign. Not yet done on this thread.
- @SirRolin — tenant/session and 24-hour entry semantics review. Not yet done on this thread.
- Owners: Nicolai and SirRolin respectively. This is their sign-off to give, not something my re-approval of the specific fix can stand in for.

**Testing performed (unchanged, still valid):**
- I verified `AND rettigheder <> '0'` in shopLookup.php's session query matches the existing convention at mysale.php:164 (not a new pattern), and that the new test (`tests/test_mysale_shop_lookup.php`) covers the exact shadow-row case that was broken, with a negative control confirming the test fails without the fix.
- I did not test tenant/session boundary behavior or the 24-hour entry window myself — that's specifically what SirRolin's requested review covers, and I'm not substituting for it.
- Earlier-flagged item (unsigned/unbound raw id links) remains explicitly out of scope per the PR description, not a regression — separate follow-up ticket, not a merge blocker.

Will re-approve (or defer entirely to Nicolai/SirRolin, whichever is cleaner) once both of their reviews land.

PR #620, opened 2026-09-15
SST-786: add CSV export covering all pages for Åbne poster report, si…

ZaynSaul:
## Summary
In Debitor → Rapporter → Åbne poster, clicking "Vis alle poster" only changed which posts
were included in the SQL (dropped the `udlignet != '1'` filter, so settled posts show too)
— it never touched pagination. The page size/`LIMIT`/`OFFSET` logic ran completely
independently of that flag. So a customer with more accounts than one page (default
100/page) still only ever saw "Side 1 af X," with no way to see a grand total to reconcile
against — exactly MEDSHOP's complaint.

Confirmed on master by seeding enough test accounts locally to trigger pagination and
hitting the report directly: `vis_alle_poster=on` still showed "Side 1 af 3 / Viser 1-25 af
68."

Rather than remove pagination from the on-screen report (that pagination exists specifically
for performance on large tenants — removing it would just reintroduce the problem it was
built to solve), added a separate **"Eksporter CSV (alle sider)"** export that returns every
matching account in one file, bypassing pagination but still fetching in bounded batches
(200 accounts at a time) so it doesn't blow up memory on a huge tenant. It reuses the exact
same filtering/aggregation logic the on-screen report uses, so the export always agrees with
what the report would show across all its pages.

### includes/reportFunc/showOpenPosts.php
- Extracted the report's account/post filtering SQL into a new shared function
  `openpost_account_query_parts()`, so the on-screen report and the export can't drift apart
  from each other over time.
- Added `openpost_export_csv()`, which performs the batched CSV export (200 accounts per
  batch).
- Added the "Eksporter CSV (alle sider)" link next to the account search box.

### debitor/rapport.php
- Added the entry point: when the export's GET param is present, buffers and discards the
  page-shell HTML that `includes/online.php` prints unconditionally, then calls
  `openpost_export_csv()` and exits — so the response is a clean CSV file instead of HTML
  wrapped around it.

### importfiler/tekster.csv
- Added two translation rows (DA/EN/NO) for the new link's label and tooltip.

## Files Changed
- includes/reportFunc/showOpenPosts.php
- debitor/rapport.php
- importfiler/tekster.csv

## How to Test
1. Seed enough test accounts (or use an existing tenant) so that Debitor → Rapporter → Åbne
   poster paginates (more than 100 accounts by default).
2. Confirm the underlying pagination bug is still present in the on-screen report itself:
   toggling "Vis alle poster" changes which posts are included but does not change
   "Side X af Y" — this is expected and unchanged by this fix (see design rationale above).
3. Click the new "Eksporter CSV (alle sider)" link.
4. Verify the response is a clean CSV file (not HTML, no page-shell content leaking into the
   file).
5. Verify the CSV contains every matching account across all pages, not just the current
   page.
6. Sum the CSV's amounts and compare against the on-screen report's own "I alt" footer total
   — confirm they match exactly.
7. Toggle "Vis alle poster" on and off, and re-export each time — confirm the CSV correctly
   reflects the same filter state as the on-screen report at export time.
8. Test on a tenant large enough to require multiple 200-account batches internally — confirm
   the export completes without memory issues or truncation.
9. Confirm the new link's label and tooltip display correctly in Danish, English, and
   Norwegian.

## Acceptance Criteria
|---|---|---|
| 1 | Tenant with more accounts than one page | "Eksporter CSV (alle sider)" link is available |
| 2 | Click export link | Returns a clean CSV file covering all matching accounts, not just the current page |
| 3 | CSV total vs. on-screen report's "I alt" footer | Match exactly |
| 4 | "Vis alle poster" toggled on/off | CSV export reflects the same filter state as the on-screen report |
| 5 | Large tenant (multiple internal batches) | Export completes without memory issues |
| 6 | Link label/tooltip | Correctly translated in DA/EN/NO |

## Verification
- Verified via direct HTTP/DB checks.
- Additionally verified by driving it through a browser logged
  in, switched tenant, clicked the real link, and confirmed the downloaded CSV's total
  matches the report's own "I alt" footer.

## Summary by CodeRabbit

- **New Features**
  - Added CSV export for open-post reports on both debtor and creditor reports.
  - Exports preserve selected date range, account range, visibility, PBS, debit/credit, and aging-bucket filters.
  - Added an “Eksporter CSV” option that preserves the current search criteria.
  - Exported reports include matching account details and totals.
  - Large exports are processed reliably without skipping or duplicating accounts.

- **Bug Fixes**
  - Improved handling of foreign-currency open items with ambiguous exchange-rate values.

- **Security**
  - Protected exported account names and numbers from spreadsheet formula injection.


LuiHedlund02:
@ZaynSaul before @SirRolin reviews: `vis_aabne_poster()` renders the CSV link for `kreditor/rapport.php` too, but only `debitor/rapport.php` got the `openpost_csv` handler (and it hardcodes 'D'), so the creditor link returns the HTML page. The export ignores aging bucket/order, so its total can differ from the on-screen total the user reconciles against. CodeRabbit's HAVING-rate finding (`showOpenPosts.php:408-418`) is open, and `firmanavn` goes to `fputcsv` unescaped.
Lead reviewer: @SirRolin (already requested). Reason: the CSV total is what customers reconcile their ledger against.
Focus: run the export from kreditor/Rapporter and from a bucket-filtered debitor view.
Next action: author fixes, then review. Ids 5151-5152 stay with this PR; #447 renumbers.

LuiHedlund02:
@ZaynSaul please add the `openpost_csv` handler in kreditor/rapport.php, where the link already renders, and answer CodeRabbit’s HAVING/exchange-rate finding before Nicolai’s review. Verify export totals against the selected aging bucket on both debtor and creditor screens, and protect CSV cells from spreadsheet-formula interpretation.

SirRolin (dismissed):

ZaynSaul:
@LuiHedlund02 all addressed:

- Added the openpost_csv handler to kreditor/rapport.php (mirrors debitor's buffer/discard pattern, kontoart='K'). Verified live: the export link now returns a real CSV from the creditor screen, and the normal creditor report request is unchanged.
- Export now honors the active aging-bucket filter (reads it from the request like the display path), so a bucket-filtered on-screen total matches the CSV total. Verified live on both the 30-60 and 8-30 buckets. I did *not* make it honor order_by/order_dir - re-sorting would require buffering every matching account in memory first, defeating the batched export's whole purpose, and row order doesn't affect which accounts or totals show (a spreadsheet sorts trivially). Flagging this as a deliberate scope call, not an oversight - let me know if you want it anyway.
- Fixed CodeRabbit's HAVING/exchange-rate finding: an account with an unresolved kurs=100 foreign-currency row now always bypasses the SQL prefilter and defers to the PHP-side resolution (which already re-derives the real rate per SST-672), instead of risking being silently dropped from both the report and export.
- The firmanavn-unescaped point looks like it was against an earlier revision - openpost_csv_safe_field() already wraps both kontonr and firmanavn as of b70b65a. Let me know if you're seeing it differently.

Ready for @SirRolin whenever you agree the above covers it.

SirRolin (commented):
Asking claude to implement the coderabbit change.

SirRolin (approved):

PR #621, opened 2026-09-15
Fix gray finance report page with a stale financial year

PeterRude:
Finansrapporten kunne vise en grå side, når sessionen henviste til et regnskabsår, der ikke længere fandtes. I test_48 blev år 2 valgt, selv om kun år 1 eksisterede. Den tomme kontoplan udløste en fatal PHP-fejl ved count($kontonr).

Rapportforsiden kontrollerer nu det ønskede år og bruger brugerens regnskabsår eller et eksisterende år som fallback. Manglende regnskabsår giver en forklaring, og en tom kontoplan håndteres også i kontovælgerens JavaScript-data.

Kun finans/rapport_includes/forside.php er ændret.

Validering:
- Genskabt den oprindelige count(null)-fejl med år 2 og data fra test_48.
- Efter rettelsen giver år 1 og år 2 samme komplette rapportforside uden PHP-fejl i CLI-testen mod test_48, kørt i en skrivebeskyttet transaktion. Oversættelsesopslag var erstattet i testkørslen for at undgå automatisk oprettelse af tekster.
- PHP-syntakskontrol og git diff --check bestået.

Manuelle testscenarier (endnu ikke browserafprøvet):
- Åbn finans/rapport.php med test_48 og et gammelt sessionsår 2: år 1 og dets kontoplan skal vises.
- Skift mellem to eksisterende regnskabsår: perioder og kontovalg skal følge det valgte år.
- Åbn et eksisterende år uden kontoplan: siden skal vises uden PHP- eller JavaScript-fejl.
- Åbn rapporten i et regnskab uden regnskabsår: en forklaring skal vises frem for en grå side.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved financial year selection by validating requested years and falling back to the user’s default when needed.
  - Added a clear message when no financial years are available.
  - Ensured reports use valid year and account defaults when selections are missing or outdated.
  - Improved account handling so available accounts are mapped and displayed reliably.


ZaynSaul (changes requested):
The core fix is right — the old code crashed with count(null) on an empty kontoplan, and the new array_keys($kontonr) + initialized arrays correctly address that root cause. Good catch, and the intval($regnaar) / db_escape_string($brugernavn) additions close a pre-existing latent SQL injection spot as a side benefit.

One thing to fix before merging: the new "zero financial years" early-return (forside.php ~L101-113) prints a bare paragraph and returns before the page header, menu, and footer are ever emitted — forside() is the only thing rapport.php calls to render the page, so this replaces the crash with a page that has no navigation at all. Could you route this through the normal header/menu/footer instead of the early return?

CodeRabbit's other finding (zero-account state where selectors disable but submit stays enabled) is still valid on the current diff too, though lower severity.

You already flagged "no financial year" as not yet browser-tested — please verify that scenario plus a year with an empty kontoplan reaching submit before this goes in.

LuiHedlund02:
@PeterRude address Saul’s current-head change request before the finance review: the no-financial-year response must retain normal page navigation, and the zero-account state must not allow an invalid submit. Test both empty-year and empty-chart states. Then request @ZaynSaul to re-review the fixes and @SirRolin to decide whether silently choosing the newest fiscal year is the intended reporting policy.

LuiHedlund02:
@ZaynSaul @PeterRude pushed `ef0e1ac` with the two review fixes.

**1. No financial year keeps its navigation.** The header/menu emission (`topline_settings.php` + the `$menu == 'T'` / `'S'` / old-design blocks and the body `<td>`/table opening) is now moved above the `!$antal_regnaar` guard, so that state renders through the exact same topline, menu and footer as the normal path — only the body area differs, showing "Der er ikke oprettet et regnskabsår. Opret et regnskabsår under Indstillinger." The guard then closes the same tables the normal path closes and includes `topmenu/footer.php` / `oldDesign/footer.php` before returning. Nothing between the old and new header position emits output, so the move is output-order safe.

**2. Empty chart of accounts blocks submit.** New `$kontiDisabled` flag (set next to `$antal_konti`) disables `konto_fra`, `konto_til` and the OK submit button together, in both the top-menu and old-design layouts. The `0 : ` fallback options are no longer emitted when there are no accounts, so PHP and `finans_forside.js` (`const konti = []`) now show the same explicit no-accounts state, plus a short Danish line explaining why OK is greyed out. `kontrolspor`/`bankReconcile` are left enabled — they do not take account bounds.

**Verified:** `php -l` clean; both states rendered offline through a throwaway harness (stubbed `db_select`/`findtekst`) against `$menu = 'S'` and `'T'`: header + menu + footer present in the zero-year output, `select name=konto_fra disabled` / `konto_til disabled` / `name='submit' disabled` present in the empty-kontoplan output, and no PHP warnings or notices in any of the four combinations.

**Not verified:** no real browser run against a live tenant — I have no tenant with zero financial years or an empty kontoplan in the local stack, so the two states were not clicked through, and the full submit path in `rapport.php` with empty account bounds was not exercised (the disabled button blocks it from the UI, but a crafted POST still reaches the handler). Worth one manual check before merge.

ZaynSaul (approved):
Verified the fix directly. The header/menu/footer include block is now emitted before the `!$antal_regnaar` early return, so the no-financial-year page keeps full navigation and footer instead of the bare message. `$kontiDisabled` correctly greys out both account selects and both submit buttons when `$antal_konti` is 0, with the new "no accounts in this chart" message. Matches the two review points exactly. LGTM.

ZaynSaul (changes requested):
Correcting my review: two issues with my approval.

**Decision:** Request changes.

1. **Outstanding named ask, not flagged.** Lui's comment (2026-09-16T08:28:56Z) explicitly asked @SirRolin to decide whether silently choosing the newest fiscal year is the intended reporting policy, not just a display bug. SirRolin has not commented anywhere on this thread. My approval ("Matches the two review points exactly. LGTM.") didn't mention this is still unanswered — a policy question I can't clear on my own.

2. **Blended testing claim.** Lui's own fix-announcement comment (2026-09-17T16:47:11Z, six minutes before my approval) says explicitly: "Not verified — no real browser run against a live tenant," and that his own test was an offline stubbed harness. My approval said "Verified the fix directly" without distinguishing my static code read from that still-open live-verification gap.

**Testing, split explicitly:**
- I ran/checked myself: read the current head (`ef0e1ac1`) directly — the header/menu/footer now render before the `!$antal_regnaar` guard, and `$kontiDisabled` correctly greys out both the year selects and submit. That part of the fix is genuinely present.
- Author-reported: Lui's fix used an offline stubbed test harness, not a live tenant.
- Never verified by anyone: an actual browser run against a real tenant with a stale financial year — Lui flagged this as still needed.

**Owner:** SirRolin for the policy question (blocks merge until answered — this determines whether "pick the newest year" is even the right behavior, not just whether the greying-out UI works); author or reviewer for the live-tenant browser check before this is treated as deployment-ready.

Will re-approve once SirRolin answers and the live check is done.

PR #622, opened 2026-09-15
Voucher window: save typed line values on Attach and unify Save/Attac…

sawaneh1:
…h field updates

## Problem

Customer reports of 14 and 15 Sept on the cash-journal voucher window (`documents.php?source=kassekladde`):

1. Values typed into the line at the top are not carried into the journal when clicking "Insert selected" (Vedhæft valgte). The file is attached but the line keeps its old values, and a new line gets the file's data instead of what was typed.
2. "Transfer data" (Overfør data) only worked when the line's checkbox was cleared. Already fixed on master in 692026bc (#606) but never deployed to the customer; this PR ships it together with the rest.

Root cause of 1: the top-row inputs were renamed from `newEntry*` / `existingEntry*` to `row_<id>_*` when multi-line vouchers were introduced, but `chooseMultipleBilag()` in `docPool.php` still read the old ids, so nothing typed was sent. Only date, invoice no. and amount were even attempted, and only for existing lines.

A code review of the same area found that `insertDoc.php` had two drifted copies of the field-update logic (Save via `action=updateOnly`, Attach via `insertFile`), an unbounded `print_r($_REQUEST)` dump to `/tmp/debug_insert.log` on every call, and `$_POST` reads without `isset`.

## Changes

`includes/docsIncludes/docPool.php`
- Attach now resolves the line to save (first checked line, else the line the window was opened for, else the new line) and sends every field of it via the existing `_collectRow()` used by Save: date, invoice no., description, debit, credit, amount, department, project, currency, VAT-exempt, due date, voucher no.
- File data is only appended for fields the user left empty.
- Server side, all three file-data sources (JS values, `pool_files`, `.info`) now only fill a field when the posted value is empty, and only on a new line. Without this the `pool_files`/`.info` fallbacks would still have overwritten a typed date or amount.

`includes/docsIncludes/insertDoc.php`
- New `insertDocUpdateKassekladdeLine($sourceId)` holds the field updates once; both Save and Attach call it. Same SQL per field as the previous Save path, same `d_type`/`k_type` prefix handling and currency lookup. Guarded with `function_exists` because the file is included once per pool file in the multi-file loop.
- Attach therefore now also stores department, project, VAT-exempt, due date, employee and voucher no., with the same empty-value rules as Save (empty debit/credit resets to 0 / `F`).
- New line on Attach takes its date in this order: typed, document date, date of an existing line with the same voucher no., today.
- Removed the `/tmp/debug_insert.log` dump.
- `$_POST` reads in the edited blocks go through `ifset()`; the one-letter account prefix is escaped before interpolation.

Not touched: `docsIncludes/insertDoc.php` at the repo root is a stale duplicate that nothing includes. No buttons, names or flow changed.

## Testing on ssl12 (Postgres)

- Existing line: typed date, invoice no., description, debit, credit, amount all saved on Insert selected, file attached. Before the change the line came back unchanged.
- New line with only description and amount typed: both kept, date taken from the file.
- `/tmp/debug_insert.log` no longer created.
- `php -l` on both files; the rendered JavaScript of both script blocks passes `node --check`; the shared function reproduces the old Save-path SQL field for field in a stubbed run.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved user-entered values when attaching pool files; imported details now fill only blank fields.
  - Improved consistency when saving or attaching documents, including voucher details, amounts, dates, descriptions, and accounting information.
  - New attachment lines now use the document date when available, followed by the existing voucher date or today’s date.
  - Removed unnecessary request debug output.


SirRolin (approved):

SirRolin:
@coderabbitai re-review

ZaynSaul (approved):
Reviewed the Save/Attach unification fix.

Both CodeRabbit findings are resolved on the current head (98865c9):
- The "0"-treated-as-empty bug is fixed via the new $postedBlank() helper in docPool.php and the trim(...) !== '' check in insertDoc.php's shared function.
- The multi-checked-line save bug is fixed via the new otherCheckedRowIds pre-save loop in chooseMultipleBilag().

Two smaller things worth a look, non-blocking:
1. Edge case: if the new (unsaved) row and an existing row are both checked, and the existing row is first in DOM order, the new row's typed values are excluded from both the pre-save loop and the attach payload — silently dropped. Worth confirming this combination can't happen in the UI, or handling it too.
2. The description says Attach now also stores medarb (employee), but neither _collectRow() nor the new typedFields object populate/send it — harmless no-op, but the description overclaims.

Since this touches kassekladde amount/date/account posting, please live-test a typed 0 debit/credit through both Save and Attach, and the new-row + checked-existing-line combo, before merging.

PR #623, opened 2026-09-16
SST-784: wrap print text by characters, not bytes

momocoder14:
Answers the wrapping half of the ticket, which acceptance criterion 2 says must not be left unresolved: the report is a reproduced code defect. Not the customer's template, and not - at least not only - manually entered line breaks.

ombryd() wraps with wordwrap($tekst, $laengde, "\n", true). wordwrap() counts bytes. The text reaching it is still UTF-8 (utf8_iso8859() runs later, inside skriv(), at the PostScript write), so every ae/oe/aa costs two of the column budget and the line breaks early by exactly its number of non-ASCII characters. Danish descriptions are the normal case, so this is systematic rather than an edge case.

Reproduced on master with synthetic text - no customer data - at the width the unaffected line in the report achieved:

  'Tjek for evt. utaetheder paa slanger foer naeste service' (with the real
  letters), width 44:

    wordwrap()     breaks after 38 characters - 48 bytes
    mb_wordwrap()  breaks after 44 characters

  Six characters of usable width lost on one line, which is the reported
  symptom: text ombrydes "foer der reelt er brugt den plads udskriften har til
  raadighed".

Adds includes/stdFunc/mbWordwrap.php: same signature and same line-breaking rules as wordwrap(), measured in characters. Pure-ASCII input is handed straight to wordwrap(), so no existing non-Danish printout can shift - verified by a differential run of 40,000 generated ASCII cases across widths 1/3/5/12/40 and both cut modes, with zero mismatches. Without mbstring it also falls back to wordwrap() rather than fataling.

All four wrap sites in the print path move together, which matters more than it looks:

  :431  ombryd(), the order line itself
  :461  ombryd(), the lokation suffix
  :2125 the measuring pass that counts description lines to decide whether an
        order line spills to the next page
  :2158 the varenr column

:2125 is the reason this is not a one-line change. It measures with the same width the renderer uses; if it kept counting bytes while the renderer counted characters it would predict more lines than get drawn and break pages in the wrong place.

Deliberately not touched: the parentheses defect, which is PR #604's, and the form template widths, which are per-customer configuration and not a code change.

Tests: tests/characterization/includes/stdFunc/MbWordwrapCharacterizationTest.test.php
- 55 cases pinning both properties: byte-identical to wordwrap() on ASCII, and no line ever exceeding the width in characters on Danish text. The 38-character break is asserted explicitly so the defect cannot return unnoticed.

Suite: 335 tests, 1 error, 7 failures - the same pre-existing NewlabelCharacterizationTest pg_query error and UsdecimalCharacterizationTest failures master produces on its own.

Verified against the real template width rather than a guessed one: the standard Dansk form in saldi_2 has formularer.xb = 52 for the beskrivelse column, and at that width the reported style of line breaks at 44 characters on master and 52 here. The measuring pass at :2125 and the rendering pass agree on the line count for every case tried, so page breaks do not move.

How much width the defect costs depends on how Danish the line is, and it can be a lot: a description made mostly of ae/oe/aa breaks at 27 characters on master against 51 here - 24 characters of the column lost. An ASCII-only line is unchanged.

That is enough to explain the report on its own. The ~13-character gap between the two lines in the screenshots needs no second cause, and an earlier draft of this message was wrong to suggest manual line breaks were likely also in play. Confirming with Auto-Lauge is still worth doing for certainty, but it is no longer a competing explanation. Their actual text is customer data and was not used here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text wrapping for UTF-8 and accented characters in generated forms.
  * Ensured character widths are calculated correctly, reducing unexpected line breaks and page-layout issues.
  * Preserved existing line breaks and behavior for standard ASCII text.
  * Improved wrapping consistency for form text, descriptions, locations, and item numbers.

* **Tests**
  * Added coverage for accented text, line-length limits, existing line breaks, long words, and compatibility with previous ASCII output.


LuiHedlund02:
Lead reviewer: @ZaynSaul
Reason: same ticket and same file (`includes/formfunk.php`) as #604, approved this morning.
Focus: `mb_wordwrap` parity with `wordwrap($s, $n, "\n", true)` on long unbreakable tokens and existing manual line breaks; `utf8_iso8859()` still runs later inside `skriv()`.
Next action: @momocoder14 rebase on top of #604 once it lands (Lui rebases #604 first); review on the diff now.

LuiHedlund02:
@momocoder14 #604 has been rebased and pushed for Saul’s re-approval. Please rebase this character-wrapping change on top of #604 (or master after it merges): same SST-784 ticket and `includes/formfunk.php`. Keep PostScript escaping at the output boundary and re-run wrapping plus print regressions after resolving the overlap.

ZaynSaul (dismissed):
Reviewed the byte-vs-character wordwrap fix. Grepped the whole repo for wordwrap( call sites - all 4 in formfunk.php (ombryd() and formularprint()) are converted to mb_wordwrap(), no sibling left byte-based, which is the incomplete-fix pattern that's bitten this area before.

Checked the interaction with PR #604 (PostScript escaping, still open/unmerged, not landed yet despite my earlier note treating it as such - flagging that correction here too): ombryd() wraps raw text and calls skriv() per line, and escaping happens inside skriv() after wrapping in both today's code and #604's branch. So mb_wordwrap's character-count decision is never made against already-escaped (backslash-inflated) text in either world - no conflict. Only a trivial header-comment merge overlap to expect between the two branches.

mb_wordwrap()'s ASCII fast path guarantees byte-identical output to wordwrap() for non-Danish text, and the word-cut/newline handling looks correct on read-through. Well-tested (55 characterization cases). Approved.

LuiHedlund02:

ZaynSaul (changes requested):
All 4 wordwrap() call sites in formfunk.php are correctly converted, the pure-ASCII short-circuit is safe, and the mbstring-missing fallback is fine. But mb_wordwrap()'s cut-loop doesn't reproduce wordwrap()'s actual long-word behavior, and I reproduced this directly rather than just reading it:

```php
wordwrap("ab løoooongword", 5, "\n", true)    // real:  ab | løoo | oongw | ord
mb_wordwrap("ab løoooongword", 5, "\n", true) // this: ab lø | oooon | gword
```

Real wordwrap() always flushes the current line and starts chopping an over-long word from a blank line — it never packs part of the long word into whatever space is left on the current line. This function's loop computes `$plads = $width - (current line length + 1)` and packs the first chunk into that leftover space instead, producing different line breaks than genuine wordwrap() for a too-long word that follows other text on the same line. It doesn't overflow the column width and it doesn't reintroduce the original byte-vs-character defect, but it contradicts the PR's own stated goal ("mirrors wordwrap()'s behaviour... without changing anything but the unit of measurement") and is exactly the parity case Lui's assignment comment asked to be checked. The characterization test's 'single long word' case never catches it because that word is always the first token (empty $linje going in).

Fix: when entering the cut branch, if `$linje !== ''`, flush it (`$ud[] = $linje; $linje = '';`) before computing `$plads`, so a too-long word always starts chopping from a blank line.

Non-blocking: mb_strlen() counts codepoints not grapheme clusters, but Danish æøå are normally stored precomposed (NFC) so this is theoretical only.

ZaynSaul (approved):
Re-reviewing after two rounds of fixes since my 09-17 changes-requested. Rather than just reading the new commit's claims, I ran the code myself:

**My over-long-word-flush finding (09-17):** confirmed fixed. Built an ASCII string with the exact same word/space-length shape as my Danish repro (`"æb løoooongword"` → shape-equivalent `"ab xxxxxxxxxxxx"`), ran both through their respective wrap functions, and the line-length shapes match exactly (2,5,5,2) with the short word correctly flushed onto its own line before the long word gets chopped.

**CodeRabbit's leading/double-space finding:** confirmed fixed — `mb_wordwrap("Æ\n  Andet", 40, "\n", true)` now correctly preserves both spaces after the hard break.

The author didn't patch either bug in place — they replaced the whole approach (three failed attempts at reimplementing wordwrap()'s rules, per the commit message) with a placeholder-byte delegation to the real `wordwrap()`, which is the right call: it makes every line-breaking decision wordwrap()'s own rather than something to keep re-deriving and re-breaking.

**Independently verified, not just re-reading their numbers:**
- Ran their own `MbWordwrapCharacterizationTest` for real (via a temp worktree + the repo's actual PHPUnit config): 62 tests, 5501 assertions, all green — matches their claim exactly.
- Wrote my own differential fuzz test (separate from theirs): 200,000 random ASCII strings with words, multi-space runs, and hard breaks, widths 1–75, both cut modes, comparing their multibyte code path (forced past the ASCII short-circuit) against real `wordwrap()` as the oracle. Zero mismatches.
- Checked double-hard-break (`\n\n`) paragraph-gap handling: preserved correctly, not collapsed.
- Re-checked all 4 `wordwrap()` call sites in `formfunk.php` are converted (render × 2, the page-break height measurement, the varenr column) and the file-history header block is updated in both touched files, per convention.
- Confirmed no merge conflict against current master.

**Not independently re-run by me:** the real-PDF-through-Ghostscript check and the live template-width verification against `formularer.xb`. Those are the author's own claims, not verified here — worth confirming if this touches production print output before the next release.

Approving on the strength of the logic-level verification above.

PR #527, opened 2026-08-28
SST-746 (skeleton): e-invoice discount-to-allowance splitter, pending EasyUBL schema answer

LuiHedlund02:
## What are the changes about?

Skeleton for the SST-746 fix (MEDSHOP: discounts must go out as Allowance, not negative invoice lines — Peppol BR-27). This PR adds the pure splitter + tests so the actual fix is a same-day wiring job once EasyUBL answers the schema question; it deliberately does **not** touch `sendInvoice()` yet.

- `includes/einvoiceAllowance.php`: `einvoice_split_allowances()` — splits built invoice lines **by sign of lineAmount** (not varenr — the DB has 4,249 negative lines across several item numbers, 224 with none): negative lines become `allowanceCharges` entries with a positive amount, `isCharge: false`, reasonCode 95, and the **goods' VAT rate** (the discount line's own momssats is 0, but the order's VAT is computed after discount — a 0% allowance fails BR-CO-10/13).
- `tests/einvoice/EinvoiceAllowanceSplitTest.php`: reproduces the rejected MEDSHOP invoice 236656 (ordre 159417): 17.196 + 0 + (−600) + 49 → 3 lines + one 600,00 allowance; totals reconcile to `sum = 16.645` / `moms = 4.161,25`; discount-free payloads byte-identical; credit-note rebates stay allowances (documents the `abs()` over-credit bug the wiring must fix).
- `phpunit.xml`: new `einvoice` testsuite.

## Gated on (before un-drafting)
1. EasyUBL's answer: does `isAllowanceCharge: true` with a negative amount get transformed server-side (then a3220aba may suffice and this simplifies), and which field carries VAT on an allowance? (Mail drafted; alternatively resending faktura 236656 after Friday's deploy answers question 1 empirically.)
2. Wiring in `sendInvoice()` must run **before** the credit-note `abs()` at the top of the ordrelinjer loop, and must supersede the a3220aba flag for negative lines — one mechanism, not two.

## How I verified this
`php tools/php/phpw.php vendor/bin/phpunit --testsuite einvoice` → OK (4 tests, 16 assertions). `php -l` clean. No runtime behavior changes in this PR — nothing includes the new helper yet.

Related: SST-746 (Jira has the full verification of master's current partial state).

## Have you checked the following?
- [x] I have checked that i am not linking to any external resources such as external style- or JavaScript-files, that could compromise stability
- [x] I have read and understood the developer guidelines

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #505, opened 2026-08-25
SD-600: unify order-creation implementations + sync next with master

ZaynSaul:
## Summary
- SD-600: characterized then unified the 4 duplicated "INSERT INTO ordrer" implementations (OrdrePage, Classic, Webshop, JWT REST) behind one shared module, `includes/order_creation.php`.
- Fixed 4 real bugs found along the way: `tilbudnr` built as an invalid hyphenated string, a missing `gruppe` default causing a Postgres error, JWT's own unscoped/unlocked order-numbering, and JWT hardcoding new suppliers as debtors (`art='D'`) even for creditor orders.
- **This branch also brings `next` up to date with `master`**: `next` was branched off an older `master` and is 60 commits behind current `master` (this branch already merged `master` into itself earlier), so this PR's diff includes that catch-up in addition to the SD-600 work itself. Per Lui: merge this into `next` now, then focus separately on stabilizing `master`.

## SD-600 commits (in order)
- `SD-600: fix tilbudnr and gruppe bugs found during order-creation characterization`
- `SD-600: add shared order-creation module (includes/order_creation.php)`
- `SD-600: migrate OrdrePage's "New order" insert to order_creation_create()`
- `SD-600: migrate Classic's opret_ordre()/opret_ordre_kopi() inserts`
- `SD-600: migrate Webshop's order insert to order_creation_create()`
- `SD-600: migrate JWT REST order creation, fixing 2 real bugs along the way`
- `SD-600: add cross-check suite proving all 4 order-creation paths share numbering`

## Test plan
- [x] `--filter OrderCreation`: 27/27 passing against a real Postgres tenant
- [x] Full characterization suite: 100/109 (9 failures are pre-existing, unrelated `finans/`/SD-601 failures, confirmed present before this branch too)
- [x] Both JWT bug fixes verified red (against pre-fix code) then green (with the fix)
- [x] `php -l` clean on every changed file

PR #624, opened 2026-09-16
Add stale PR workflow with a dry-run activation gate

LuiHedlund02:
Adds the prepared PR hygiene workflow: mark a PR stale after 30 inactive days and close after another 30; `udskudt` exempts intentionally parked work. Issues are excluded. Drafts follow the same rule unless parked.

Manual `workflow_dispatch` defaults to `dry_run=true`. Scheduled runs are gated by the repository variable `SALDI_STALE_ENABLED=true`, so merging this file cannot start closures before the first dry run has been inspected.

Validation: YAML parses and whitespace checks pass; actions/stale v11 resolves to an existing upstream tag. This does not run functional application tests.

Activation after merge: dispatch with `dry_run=true`, read the complete log and verify the proposed labels/closures and parked exemptions, then set `SALDI_STALE_ENABLED=true`. The manual dry run cannot be dispatched while this new workflow is absent from the default branch.

The September 16 queue pass parks #527, #505 and #440 explicitly. Other deferred drafts/release work should retain their existing deferral until intentionally resumed. GitHub activity includes bot comments, so this workflow does not measure human review age.


LuiHedlund02:
@ZaynSaul please review the prepared stale-PR workflow. YAML/whitespace checks pass and the upstream action tag exists. Focus on the 30+30-day thresholds, `udskudt` exemption and the activation gate: scheduled runs stay disabled until the first manual dry-run log is inspected and SALDI_STALE_ENABLED is set. The initial dry run remains blocked until this new workflow is merged to the default branch. All intentionally deferred PRs identified in the September queue have been labelled before activation.

ZaynSaul (approved):
Reviewed the stale-PR workflow. Verified the dry-run/activation gate is genuinely wired, not just documented:
- The job's `if` only runs on workflow_dispatch or when the SALDI_STALE_ENABLED repo variable is 'true' - since that variable doesn't exist yet, scheduled/cron runs stay off after merge.
- Manual dispatch defaults dry_run to true.
- No `run:` shell steps exist at all - the only step invokes actions/stale@v11 via `with:` inputs, so there's no untrusted-context-into-shell injection vector to worry about.
- Confirmed actions/stale@v11 resolves to a real, immutable-tagged release (v11.0.0), not a mutable floating tag.

Permissions are scoped to pull-requests/issues write only. Thresholds (30/30, udskudt exemption, drafts included) match the description. Merging this cannot itself close or label anything - approved.

PR #630, opened 2026-09-16
Preserve voucher input fields when opening document previews

LuiHedlund02:
Opening a document preview can erase the new voucher row's debit/credit accounts and other input values: the file-list link rebuilds the page URL using pool context only. A row containing debit 6400, credit 58300, invoice 750594 and amount 1.048,38 becomes blank after clicking a file. Additionally, when the same voucher already has journal lines, the automatically checked new row is hidden below those lines, so “Overfør data” appears to do nothing.

This change keeps the new row first and visible, and collects its current fields before navigating to a document preview. The page receives all new-row values, including department, project, currency, VAT-exempt and due date. Values are escaped once by the renderer so quotes and ampersands survive navigation. Existing saved rows continue to load from the journal database.

### Validation

- Reproduced preview data loss through the running PHP application and PostgreSQL: accounts/date/invoice/description/amount were populated before selecting a file and empty afterward.
- Repeated against the fix in Chromium using an isolated application checkout and disposable database: all current new-row fields survive the real file-list click, including special characters and VAT-exempt state.
- Clicked Save, reloaded, and checked PostgreSQL: debit 6400, credit 58300, amount 1048.38, invoice 750594, description, date, department, project and due date persisted.
- Attached the preview file to that saved row: the documents record links to the intended row and the accounting fields remain unchanged.
- Verified a pre-existing saved journal line loads debit 6400, credit 58300 and amount 1.048,38 from the database.
- Three committed Playwright regressions pass: all new-row fields, explicit blanks/zero replacing stale values, and saved-line navigation preserving its source ID.
- PHP-rendered browser checks also pass for the hidden-row fix: new row first with labels, transfer OK/Cancel, expand/collapse, remembered expanded state, existing selected row, and a voucher without existing rows.
- `php -l` and `git diff --check` pass. PHP reports the same pre-existing interpolation deprecations as baseline.
- No production writes. Full-page tests used a disposable local clone; ssl3 SSH was unavailable during the follow-up investigation, so the current live customer session has not been verified.

### Run the committed regressions

```sh
npx playwright test --config tests/ui/playwright.config.mjs tests/ui/voucher-preview-fields.spec.mjs
```

### Manual regression checks

- `docPool.php` / `openPoolFile()`: open a new voucher line with populated accounts, change debit/credit and description, and click two different pool files. Confirm the current values remain visible, including cleared fields and a zero amount.
- Row rendering / `toggleBilagRows()`: open “Ny linje” on a voucher with several existing lines; confirm the new checked row has labels and remains visible through expansion, collapse and reload.
- `transferDataFromSelectedFile()`: cancel then confirm a document transfer; amount/date/invoice/description update only on confirmation, while debit/credit retain their values.
- `saveAllRows()` / `_collectRow()`: save and reopen the new row, checking all entered fields. Repeat for an existing journal line.
- `chooseMultipleBilag()` / `insertDoc.php`: attach to a new row and an existing selected row, checking the resulting journal values and attachment target.

## Summary by CodeRabbit

- **Bug Fixes**
  - Unsaved voucher fields—including department, project, currency, VAT status, and due date—are now preserved when reopening a document preview.
  - Explicitly cleared or zero-valued fields remain intact during preview navigation.
  - Navigating between pool-file rows now carries the correct voucher details with it.
  - New voucher rows remain visible at the top of the preview when no source line is selected, while existing rows are collapsed for clearer viewing.


ZaynSaul (approved):
Verified the core of this myself: removing `htmlspecialchars()` at the `$displayX` assignment site isn't a regression — `renderBilagRow()` (unchanged here) already escapes every field before printing into `value="..."`, so this correctly eliminates a double-escaping bug rather than introducing an XSS gap. Traced the momsfri checkbox → "0"/"1" string → PHP `empty()` round-trip for both checked/unchecked states — correct. The `.set()` vs `.append()` duplicate-param risk is explicitly covered by the second test.

CodeRabbit's note about the test using a synthetic `bilagEntry_new` fixture instead of clicking the real rendered pool-file row is a fair nitpick, but not blocking given the PR body documents separate real-browser Playwright runs and manual PHP-rendered checks. Worth a follow-up if you have time, not required for merge.

LGTM.

PR #636, opened 2026-09-17
MB-58: Tilbage efter udskrift fra ordren går til ordren igen, ikke ordrelisten

LuiHedlund02:

Havemøbelland (via Thor, 17/9): efter fakturering vises ordren klar til udskrift, og "Tilbage" sender brugeren til ordreoversigten i stedet for tilbage til den fakturerede ordre. Det koster et ekstra opslag pr. ordre, fordi følgesedlen printes fra ordren.

Årsag i `includes/udskriv.php`: DO-grenen (LOE 20260217/20260428) sender til `../debitor/ordreliste.php`, medmindre filen `temp/<db>/area<bruger_id>.txt` indeholder "faktura". Ingen kode på master skriver den fil, så alle udskrifter startet fra ordren endte på listen. Desuden var `$art == ('DO' || 'PO')` altid sand.

Rettelse:

- "Tilbage" efter en udskrift startet fra ordren går igen til `../debitor/ordre.php?tjek=$id&id=$id&returside=…` (som før februar 2026).
- Masseudskrift fra ordrelisten (`locat=1` sat af `ordreliste.php`) går fortsat til listen.
- Den døde area-fil-kontrol er fjernet, og art-sammenligningen bruger `in_array()`.

Kun `includes/udskriv.php` ændres. Samme kodesti som SST-772 (MEDSHOP, "Luk" efter følgeseddel ved dellevering).

## Affected functions/files and manual test scenarios

- `includes/udskriv.php`, `$menu == 'S'`-grenen der tegner "Tilbage" efter PDF-udskrift.
  - Åbn en ordre, tryk Fakturér (eller Udskriv med "Udskriv til" = PDF). Tryk "Tilbage" på udskriftssiden: ordren åbnes igen, ikke ordrelisten.
  - Ordreliste → markér flere ordrer → Print følgesedler/plukliste: "Tilbage" går til ordrelisten som hidtil.
  - Kreditorordre-udskrift og rykker-udskrift: uændret retur (kreditor/ordre.php hhv. rykker.php).
- Verificeret lokalt (docker, test_36, ordre 122484): før = `../debitor/ordreliste.php`, efter = `../debitor/ordre.php?tjek=147064&id=147064&returside=…`, og linket åbner ordren. `php -l` ren.

Jira: MB-58 (relates SST-772).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved return navigation after printing orders.
  * Orders now return to the order list only when the configured location and return path match; otherwise, users return to the specific debtor order.
  * Removed outdated routing behavior based on invoice status and area-file checks.


LuiHedlund02:
Lead reviewer: @ZaynSaul
Reason: Post-print "Tilbage" navigation in `includes/udskriv.php` (Havemøbelland, MB-58); isolated fix, removes a dead area-file check.
Focus: Bulk print from ordreliste (`locat=1`) still returns to the list; SST-772 partial-delivery path unchanged; kreditor and rykker branches untouched. CodeRabbit's empty-`returside` case: decide if it is reachable from `debitor/ordre.php`.
Verification: Author reports local docker test_36, order 122484, href before/after.
Next action: Review this PR.

ZaynSaul (approved):
Verified: $art == ('DO' || 'PO') was a real always-true bug (PHP evaluates the parenthesized OR to boolean true first), now correctly in_array(). Bulk print from the order list (locat=1) still goes to ordreliste.php; every other DO/PO print now returns to the order, matching the manual repro in the PR body. One note: the PR says 'no code on master writes' temp/<db>/area<bruger_id>.txt, but debitor/ordre.php:3446 does write it on every $_GET[valg] — it just never gets set to exactly 'faktura' from anywhere except this now-removed branch itself, so the practical conclusion holds, just the description is slightly off. Not blocking; that write is now fully orphaned as a minor follow-up if anyone wants to clean it up. LGTM.

ZaynSaul (changes requested):
Correcting my review: Lui's handoff explicitly asked me to decide whether CodeRabbit's empty-`returside` case is reachable from `debitor/ordre.php`. I addressed the `$art == ('DO'||'PO')` always-true bug but never answered that specific question, and approved anyway. Also, CodeRabbit's own CHANGES_REQUESTED review is still standing/undismissed — my approval treated the PR as fixed without resolving that thread.

**Decision:** Request changes.

**Unaddressed question (Lui's, still open):** In `includes/udskriv.php`, the new order-vs-list back-link logic only runs inside `if ($returside) { ... }` (~line 440 onward, under `if ($menu == 'S')`). When `$returside` is empty, there's a separate, earlier, unconditional block (`if ($valg=="tilbage" && !$bgr)`, ~line 133) that redirects straight to `../debitor/ordre.php?id=$id` regardless of where printing started. I have not confirmed whether that earlier block is the actual code path hit by the real "Tilbage" click this ticket is about, or a separate/legacy entry point (e.g. a different print trigger). If it's the former, the ticket's underlying bug (back-from-print always returning to the order instead of the list) could still reproduce outside the `$menu == 'S'` flow — this needs tracing before the fix can be called complete.

**Process gap:** CodeRabbit's CHANGES_REQUESTED review on this PR has not been dismissed or answered with evidence — my approval shouldn't have been read as superseding it while it's still open.

**Owner:** me, to trace the `$menu` value and confirm which block real "Tilbage" clicks hit — or the author, if they can confirm directly which flow corresponds to the ticket's repro steps.

Will re-review once the reachability question is resolved and CodeRabbit's thread is addressed.

includes/udskriv.php
```diff
@@ -45,6 +45,8 @@
 // 20260820 CX/PHR Return reminder prints to the reminder instead of the debtor order form.
 // 20260901 CL/LH SD-664: ret <?= i dobbelt-quoted streng (redirect ved manglende pdftk blev aldrig udfort)
 //             og giv retur-link ved 'PDF-fil ikke fundet' i stedet for blindgyde (browser-Back re-POSTer)
+// 20260917 CL/LH Tilbage after printing from an order returns to that order again, not the order list
```
ZaynSaul, line 48: History-comment collision with master's JOB-124/SST-789 lines from #605 — trivial, keep both.

includes/udskriv.php
```diff
@@ -435,31 +437,20 @@
...
-			###########
 			if ($menu == 'S') {
 				print "<table width=100% height=100%><tbody>"; 
 				if ($returside) {
 				 if ($art == 'R') {
 					$href = "../debitor/rykker.php?rykker_id=" . (int)$id;
 				 } elseif (substr($art,0,1)=='K'){
 					$href="\"../kreditor/ordre.php?tjek=$id&id=$id&returside=" . urlencode($returside) . "\" accesskey=\"L\"";
-				 }elseif ($art == ('DO' || 'PO') && (strpos($returside, "ordreliste.php") !== false) && $locat) {
+				 } elseif (in_array($art, array('DO', 'PO')) && $locat && strpos($returside, "ordreliste.php") !== false) {
```
ZaynSaul, line 447: Real conflict: master (#605) inserted a new elseif branch (`strpos($returside, '../debitor/ordre.php?') === 0`) right before this one. This branch independently fixed the always-true `$art == ('DO' || 'PO')` bug to `in_array($art, array('DO', 'PO'))`. Both edits need to survive the merge in the same relative order — keep master's new branch, then this fixed condition + comment right after it. Don't let the resolution revert to the buggy `$art == ('DO' || 'PO')` form.

ZaynSaul (changes requested):
This branch now conflicts with `master` (`CONFLICTING`/`DIRTY`) after #605 (SST-789, "Print all selected invoices in one ordered PDF") merged, which also touched `includes/udskriv.php`'s back-link logic in the same function.

Checked with `git merge-tree` — two conflicts in the same file:

1. **Top of file, history comment** — trivial: this branch's `20260917 CL/LH Tilbage after printing...` line vs master's JOB-124/SST-789 lines from #605. Keep both.
2. **The `$href` elseif chain a few lines below** — this one needs care, not just "keep both". Master (#605) *inserted a new elseif branch* right before the DO/PO check: `elseif (strpos($returside, '../debitor/ordre.php?') === 0) { ... }`. This branch, independently, *fixed* the existing DO/PO condition itself — from the always-true `$art == ('DO' || 'PO')` to the correct `in_array($art, array('DO', 'PO'))` — and added a comment. These are two different, compatible edits to adjacent lines: the fix needs to land, and master's new branch needs to stay in the chain, in the same relative order (master's new elseif before the DO/PO one, same as now). Whoever resolves this should keep master's new branch verbatim and apply this branch's `in_array` fix + comment to the DO/PO elseif right after it — not accidentally revert to the buggy `$art == ('DO' || 'PO')` condition, which is a real bug (that expression is always true regardless of `$art`).

Requesting changes to get this resolved — the review itself stands as approved; this is scoped to the conflict. Given part 2 touches actual control flow, worth a quick re-test of the print-and-back flow (both from an order and from the order list) after merging, not just an eyeball of the diff.

Loe4g:
@LuiHedlund02  There seems to be a mix-up in the attribution here, and it is the same one the SST ticket carries which I have pointed out.

f5b603d2 (https://github.com/DANOSOFT/saldi/commit/f5b603d2e4a5a47dd02a69151522a8b8a07ba560) is not where
this broke - it is where it worked. That commit added both halves of the mechanism: the area-file read in
includes/udskriv.php and the write that populated it in debitor/ordre.php ($opValue = $_GET['valg']). At that
point the redirect chose correctly: a print started from the order returned to that order, and the list /
pick-list / delivery-note context went to the order list.

What removed it is 92dd1295
(https://github.com/DANOSOFT/saldi/commit/92dd1295467445da5a6177d362ce9eff830ed949, PR #356), which deleted
exactly these two lines in debitor/ordre.php:

    $opValue = $_GET['valg'];
    file_put_contents("../temp/$db/area$bruger_id.txt", $opValue, LOCK_EX);

The reader is still in udskriv.php - what disappeared was what fed it. It came back on 2026-09-16 in 886ad6ab
(https://github.com/DANOSOFT/saldi/commit/886ad6ab609f2575dcab605be71061c8e6ce6a0b, PR #526), guarded so an
invoice save no longer blanks the value, so "nothing ever wrote it" does not hold on this PR's base.

The 20260217 half of the citation is off too: 8d0a3b005
(https://github.com/DANOSOFT/saldi/commit/8d0a3b0059d2a47e4e2f2aa3ab1b076d7ee36f5a) only re-indented that
block - same conditions, same targets. (The always-true `$art == ('DO' || 'PO')` is real, from 6f965cb2
https://github.com/DANOSOFT/saldi/commit/6f965cb2d91a1bef69622d9c06dce00b675c445a, but it only widens the bulk
branch, which also needs locat.)

One consequence of removing the check instead of repairing it: the multi-order prints pass id=-1
(ordreliste.php:1964/1966, massefakt.php:118, ret_genfakt.php:385 - only ordreliste.php:1968 sets locat=1), so
they fall into the unconditional branch and get ordre.php?tjek=-1&id=-1, which renders an empty order form
(select * from ordrer where id='-1' matches nothing). I ran this block from master and from this branch with
those inputs: master yields the order list for them, this branch yields the empty order page. Keeping the check
and keying on a real order id - id > 0 to the order, bulk to the list - fixes the case this PR is about without
that.

ZaynSaul (changes requested):
**Decision:** Request changes (in addition to the open items from my prior review).

The area-file check this PR removes isn't dead: `debitor/ordre.php:3457-3459` still writes `temp/<db>/area<bruger_id>.txt` whenever `$_GET['valg']` is present - restored by #526 on 2026-09-16, guarded so an invoice save with no `valg` doesn't blank it. That contradicts this PR's stated premise, and the removed `$value == 'faktura'` branch was the only thing sending non-`locat` bulk prints (formular=4 fakturaer from `ordreliste.php`/`massefakt.php`/`ret_genfakt.php`, which pass `id=-1` and no `locat`) back to the order list instead of the order itself. Needs a direct before/after test on that path before merge - those calls need to keep landing on the order list, not fall into the new unconditional `else` branch.

Also, the commit attribution in the PR body has it backwards: `f5b603d2` is where this worked, not where it broke - it added both the read in `udskriv.php` and the write in `ordre.php`. `92dd1295` (#356) is what deleted the write. `8d0a3b005` only re-indented the block; it didn't change conditions or targets.

**Owner:** author, to re-verify the bulk-print (`id=-1`, no `locat`) path against both master and this branch before merge.

PR #640, opened 2026-09-18
Sd 700 Keep the document pool list order, position and search while a bilag is opened or handled

Loe4g:
Opening a bilag in the document pool moved that row to the top of the list, and every action that
reloads the pool came back at the top of an unfiltered list.

- `renderFiles()` no longer puts the document that is open in the preview pane into its own
  `activeRows` bucket. That bucket was emitted as the first `<tbody>` row, so the clicked row jumped to
  position 1 and the previously first row was pushed down — the list reordered itself under the user on
  every click. The row keeps its highlight and is now revealed with `scrollIntoView` when it is
  off-screen, which is what the hoist used to guarantee.
- The list the user was looking at survives those reloads: the scroll offset of `#fileListContainer`
  and the search text are stored per tab and per tenant in `sessionStorage`, re-applied before and after
  the client-side render. That happens before every navigation or reload the pool performs (row click,
  assign number, single and bulk delete, extract, extract all, upload, kassekladde line
  save/save-all/duplicate).
- The column sort is deliberately **not** kept: it re-sorted the list after an edit had changed the
  sorted field, so a row the user had just worked on came back somewhere else. It stays session-only,
  and while it is active its header is marked (red background, white bold text, ▲ ascending / ▼
  descending, neutral ⇵ when unsorted).
- No SQL added or changed, and this PR writes nothing to the database. The document is identified by the
  filename the URL already carries as `poolFile`, which is also the row's `data-pool-file` key and
  UNIQUE in `pool_files`.
- Sort semantics unchanged: `_docPoolData.php`'s `file_date DESC, updated DESC`, the amount/date match
  buckets and the column comparator are untouched.

Ticket: SD-700 (support case SST-798).

## Scope note (short, vs the ticket's written scope)
- State is keyed by tenant (`docPoolList_<db>`), not "regnskab + kladde": the pool is per tenant and
  `kladde_id` is usually empty on the standalone pool URL.
- The row is identified by filename, not `pool_files.id`: `_docPoolData.php` does not return `id`, and
  the ticket asks for no new SQL.
- "Prefer updating the row in place over reloading" is not in this PR: the preview pane is rendered
  server-side, and for bmp/jpg/jpeg/png/tif/tiff that path converts the file to PDF, deletes the
  original and renames the pool file, so a JS swap would skip it. Proposed as a follow-up ticket.
- Edit-info already saves in place (no reload) and tick marks already survived, so neither needed a
  change — the ticket's "every action reloads" only holds for the other triggers.
- Removing the hoist does change the rendered order; that is the fix, and the sort rules themselves are
  unchanged.

## Known pre-existing behaviour while testing (not from this PR)
Rows that were scanned or edited collect at the bottom of the list, and editing a row moves it down
there. Cause: `pool_files.file_date` is a varchar holding two shapes — a timestamp from when the
document entered the pool, and a date-only value written by the date normaliser the scan/save path uses.
Sorted as text, every date-only row lands below every timestamped one; and the pool-row edit sends back
only the date part, so one edit converts a timestamp into a date-only value and the row drops into that
lower group.

**Slight Addition:** 

If a column header is red, a sort is active (this PR marks which column); if none is red, the order you
see is the pool's own `file_date` order. Logged for a follow-up ticket.

**Also pre-existing:** a row showing `Unknown` after a scan is the extractor's placeholder stored as the
subject, not a renamed file (the currency equivalent makes the API answer HTTP 500, so those scans save nothing, perhaps because not on prod).

P.S: other tickets are to follow this to handle other UseCases as we proceed.

## How I verified this

Automated:
- `php -l includes/docsIncludes/docPool.php` clean. Because this script is printed from a PHP heredoc,
  the **served** page is checked as well: HTTP 200, ~162 KB, no Fatal/Parse/Warning output, and the pool
  script extracted from that response passes `node --check`.
- Headless Chrome against the real render path (`_docPoolData.php` stubbed, four rows in
  `file_date DESC, updated DESC` order, URL pointing at the last row): with only the removed hoist put
  back the clicked row renders first; with this branch the order is unchanged and the clicked row is the
  selected one. In the same setup the search text survives a reload while the amount sort does not, a
  sort planted in the storage by an older build is ignored, and scroll offset 400 returns as 400.
- Five committed Playwright regressions in `tests/ui/pool-list-order.spec.mjs`
  (`npx playwright test --config tests/ui/playwright.config.mjs`): the open document keeps its place;
  search survives a reload but the sort does not; an older build's stored sort is ignored; the sorted
  column is marked with the right direction; the scroll offset survives. The first fails without the fix.
  Playwright is not installed in the environment I used, so these assertions were executed in headless
  Chrome instead of through the runner.
- No PHP logic changed, so the PHPUnit suite was not run for this PR.

Manual, in Prodtest on ssl12 / test_37, one pool with 50+ documents:
- Reproduced on master: clicking a row in the pool list scrolled the list back to the top and moved the
  clicked row to position 1, so the row the user was reading moved away.
- Repeated against this branch: the list keeps its position — the clicked row stays where it is, is
  highlighted, and its document opens in the preview pane; editing a row no longer pushes it to the top.
- With a search filter and a column sort applied while clicking through: the position and the filter are
  kept, the sorted column is marked with a red header and an arrow, and after a reload the sort is gone
  (session-only) while the filter stays.
- Also exercised with the list staying put: assign a number/insert into a kassekladde line, delete,
  extract ("Opdatér"), extract all, upload; a jpg still converts to PDF as before, and PR #630's typed
  voucher fields still survive opening a preview.

## Summary by CodeRabbit

- **User Experience**
  - Pool list order and selected position are preserved when reopening or reloading documents.
  - Search text and scroll position persist across page reloads.
  - Scroll position is restored after returning from document actions such as saving, deleting, uploading, or duplicating.

- **Sorting**
  - Column sorting remains available during the current session.
  - The active sort column is clearly indicated with a visual arrow.
  - Sort selection is not retained after a full reload.

- **Tests**
  - Added coverage for list order, search persistence, sorting indicators, and scroll restoration.


ZaynSaul (dismissed):
Full review: no correctness regressions found. Verified all 7 `window.location` reload sites call `savePoolListView()` beforehand; sort-toggle refactor confirmed behaviorally equivalent to the old code (traced the short-circuit logic); sessionStorage state confirmed tenant-scoped (`docPoolList_<db>`) and client-side only — no cross-tenant/cross-user leakage possible.

**Decision:** Approve.
**Blockers:** none.
**Testing:**
- Verified by me (read code directly): hoist removal, all reload-site coverage, sort-toggle equivalence, sessionStorage scoping.
- Author-reported, not independently reproduced: live Prodtest verification on ssl12/test_37, the master-branch repro of the reported bug.
- Unverified by anyone: whether CI actually executes `tests/ui/pool-list-order.spec.mjs` through the real Playwright runner (author notes Playwright wasn't installed locally when writing it).

**Nonblocking follow-up:** confirm the Playwright test is actually wired into CI so it doesn't silently bit-rot.

SirRolin (dismissed):

SirRolin (approved):

PR #643, opened 2026-09-18
MB-41: Enter moves the focus to a new line + description-only saves and 0-amount lines stay put

Loe4g:
Pressing Enter (or Gem) in the kassekladde left the caret where it was, and a save that created a line could
leave the caret on a row the user had not typed in. Ticket: MB-41.

- **Focus follows the new line.** A save that *creates* a line puts the caret on the new blank line's `bilag`
  field, whatever field Enter was pressed in. A save that only *updates* a line leaves the caret where it was,
  so the correction can be checked; moving on is the user's choice. The row that was saved also stays where it
  was on screen, even when the correction changes what the list is sorted by.
- **A line holding only a description now has a row to continue in.** The trailing blank row was only rendered
  when the last line carried a debit/credit, so a description-only save left no empty row — the caret stayed on
  the line just saved and the user had to click to continue.
- **A line created by a save is rendered last**, whatever the list is sorted by. With the list sorted by `Beløb`
  a line with amount 0 (and an empty bilag, stored as 0) otherwise sorts to the top, so the line just typed
  disappears from where the user is working — in a 50+ line kladde, "where did it go?". The pin covers that one
  render only: the sort is untouched and the line takes its sorted place on the next load.
- The regression came in with `e996d53f` (2026-08-20, an unreviewed live update that touched four files): it
  added a second focus setter in a `DOMContentLoaded` handler that used the focus field as printed *before* the
  row loop advances it, and it runs after the inline script that focuses the advanced field — so it pulled the
  caret back on every save. The server-side advance itself dates from the first git import (2013) and was gated
  on the amount/department fields.

## Notes for review
- No SQL structure changes and no new user input reaches a query: the ordering terms are built from `intval`ed
  row ids on the query the kladde is already rendered from, so row data, sums and `pos` stay aligned with the
  row order.
- The render that follows a save is ordered by the submitted on-screen order (`array_position()` on Postgres,
  `FIELD()` on MySQL), so the row the user edited keeps its position and a row this save created — which is not
  in that list — sorts last. The list's own sort is untouched and applies again on the next load.
- The tracker (`kk_note_new_line`) is what tells the focus and the pin that this save created a line: it records
  the id the database generated for that insert, read from the same connection (`currval()` on Postgres,
  `mysqli_insert_id()` on a MySQL install), and it runs only for inserts. `opdater()` runs once per request
  under `!$fejl`, so a validation error changes nothing.
- The `+N` / `+=` bilag shortcut inserts through `indsaet_linjer()` rather than `opdater()` and keeps its old
  behaviour — not covered here.
- The ticket's "a line in the middle of a long kladde" is read as "must also work on a long, scrolled kladde",
  not as "a mid-list save must move the caret".

### Manual, on a test tenant on ssl12 (test_37) journal7 - master branch vs this branch: the same kladde, the same user, the same sort

These are the developer's manual test runs on tenant_db (master db as the registry); the case in the last bullet
came out of the same runs after the first review, and its fix landed after that run, so only the SQL check below
covers it.

- **Enter on a completed line.** master: the line is created but the caret stays in the field Enter was pressed
  in (usually `Beløb`). this branch: the caret is in the new line's `bilag` field, so typing continues without
  the mouse. Also covers saving from `Faktura`, `Beskrivelse` and `Afd.`, which never moved the caret even
  before the regression.
- **Enter on a line holding only a description.** master: the caret stays at the same row *position* near the
  bottom (second-to-last) while the saved line is sorted away to the **first row** — an empty bilag is stored
  as 0 and the amount is 0, so it sorts ahead of everything. The user ends up with the caret on a line they did
  not type in and their new line at the top. this branch: the created line is rendered last, where it was typed,
  and the caret is on the new blank row's `bilag` field.
  Reproduce on master: put a description only in the bottom line, press Enter, then look for the line at the top
  of the list while the caret sits near the bottom.
- **A new line with amount 0 no longer jumps away.** Reproduce: sort by `Beløb`, fill the bottom line with
  amount 0, press Enter. master: the line takes its place in the amount sort (top). this branch: it stays at the
  bottom where it was typed, next to the row the user continues in; after a reload it sits in its sorted place.
- **Editing a line that has lines after it keeps the caret where it was saved**, on master and here alike.
- **Editing a middle line while the list is sorted by `Beløb`.** The manual run turned this up after the first
  review: correcting an amount sent the row to its new place in the amount sort while the caret, which is
  addressed by row position, stayed behind. Fixed in `6b618bc5` — the row stays where it was and the caret stays
  on it, and the row takes its sorted place on the next load. Reproduce on master: sort by `Beløb`, correct the
  amount on a middle line so it belongs higher up the list, and save. The fix landed after that manual run, so
  the SQL check in the Automated block below is what backs it.
- **Long kladde, after a save and on reload:** the field the page focuses is scrolled into view, so the line
  being worked on is on screen.

### Automated
- `php -l finans/kassekladde.php` clean; the served kladde page renders with no PHP warnings, and the script it
  prints still contains the focus handler that prefers `window.savedFocus`.
- The render-order change checked directly in SQL on a real kladde: the row that leads the `Beløb` sort moves
  last under `pos`, `bilag,transdate` and `amount` ordering; with the submitted on-screen order the same order
  comes back, a row absent from the list sorts last, and an edit that a plain amount sort would move to the
  front (`2,4,3,5,6`) stays in place (`4,2,3,5,6`).
- The focus and blank-row conditions extracted from the file and run against fixtures for amount-only,
  description-only, invoice-field, mid-list edit and last-line edit: caret on the new blank line for every
  creation, unchanged for updates.
- The id capture is checked for both database types: `pg_get_serial_sequence('kassekladde','id')` resolves and
  `currval` returns the value from the inserting session, and the branch that chooses between `mysqli_insert_id()`
  and `currval()` was run for `postgresql`, `postgres`, `mysqli`, `mysql` and an unset `$db_type` — MySQL only
  for the two MySQL values. The insert/update detection returns the expected answer for both query shapes, so
  only inserts record an id.
- No automated regression test is added: the kladde has no UI harness (`tests/ui` specs cover the document
  pool), so this rests on the block-level checks above plus the manual scenarios.

## Review updates (after the CodeRabbit reviews)

- **`3febb8d1` — the two findings from the first review.** The new line's id no longer comes from `SELECT
  MAX(id)`, which a concurrent insert in another session could win; it is read from the connection that ran the
  insert (`currval(pg_get_serial_sequence('kassekladde','id'))`), and only when the statement is an insert. The
  historical fallback that advanced the caret after an update is gone, and the tracker is insert-only — it used
  to fire for updates too, because the shared `if ($qtxt)` block also covers the UPDATE branch, which is what
  actually moved the caret when an existing line was edited.
- **`1741a791` — docstring coverage.** `kk_note_new_line()` and `opdater()` carried the diff without a docblock
  and the check asked for 80%; both have one now, in the style already used in the file. No behaviour change.
- **`6b618bc5` — the row the user edited no longer moves.** Removing the accidental pin on updates exposed the
  case above: an edited row took its new place in the active sort and the caret, which is positional, stayed
  behind. The render that follows a save now keeps the submitted on-screen order, so the edited row stays put
  and the caret stays on it; a row created by that save is not in that list and sorts last, where it was typed.
- **`894a10b0` — the finding from the second review.** The id capture skipped MySQL/MySQLi installs, so there
  the created line was not pinned, the focus stayed put and an amount-0 or description-only line got no blank
  row. `currval()` needs a sequence, so the branch now mirrors `db_modify()`'s own `$db_type` test and uses
  `mysqli_insert_id(db_query_connection(false))` for mysql/mysqli, keeping the `currval` query unchanged for
  Postgres; `kk_note_new_line()` runs only for an id above 0, so a failed insert or an install whose id is not
  auto-generated falls back to the behaviour without the pin. (The finding's premise — that this branch called
  a non-existent `db_insert_id()` — did not match the code, but the gap it described was real.)

## Summary by CodeRabbit

- **New Features**
  - Newly created cash-journal lines now remain at the end of the displayed list, regardless of the selected sort order.
  - Edited lines retain their submitted on-screen order after saving.
  - After saving a new line, focus automatically moves to the next blank line, including description-only entries.

- **Bug Fixes**
  - Improved focus behavior after saving cash-journal entries.
  - Failed or unsupported line creation no longer interrupts saving or focus handling.


Loe4g:
@coderabbitai review

Loe4g:
@coderabbitai full review

Loe4g:
@coderabbitai full review

Loe4g:
@coderabbitai review

ZaynSaul (approved):
Full review: all 3 CodeRabbit findings (MAX(id) race, fallback focus-advance on update, MySQL/MySQLi id capture) verified genuinely fixed on the current head — re-checked each against actual current code, not just the thread's resolved flag. No off-by-one in focus targeting, sort-branch coverage complete across all 5 kksort branches, connection atomicity confirmed (db_modify/db_select share the same connection).

**Decision:** Approve.
**Blockers:** none.
**Testing:**
- Verified by me: CodeRabbit fix resolution, focus/index logic, sort coverage, connection atomicity, DOMContentLoaded/savedFocus execution ordering.
- Author-reported, not independently reproduced: manual browser test matrix on test_37/journal7, SQL fixture checks.
- Unverified by anyone: no automated regression test exists for this file (kladde has no UI test harness); no live run performed in this review.

**Nonblocking:** none.

PR #646, opened 2026-09-19
Escape the request-derived values grid.php writes into HTML attributes

momocoder14:
CodeRabbit raised the unescaped search term on PR #447 on 2026-08-04. That PR is being closed as superseded, and the finding was never actually fixed there - the grid.php change on that branch added a renderSearch hook and left the default output exactly as it was - so it is fixed here instead, off master.

includes/grid.php interpolated four values straight from $_GET into HTML attributes:

  :1086  value='$columnSearchTerm'   $_GET['search'][$id][<field>]
  :966   value='$sort'               $_GET['sort'][$id]
  :967   value='$menu'               $_GET['menu'][$id]
  :1235  value="$offset"             $_GET['offset'][$id], never cast

The search term is the worst of the four, for two reasons. grid.php is shared by every list page, not just the order list. And the term does not only reflect: it is json_encoded into datatables.search_setup at :322/:404 and read back into $searchTerms on later requests, so a crafted value persists for that user until the search is cleared. Reflected and stored, from one crafted link.

Demonstrated against the real render_table_headers() before fixing. With the search term set to

  ' onfocus=alert(document.domain) autofocus x='

the emitted markup was

  <input ... value='' onfocus=alert(document.domain) autofocus x='' placeholder=''>

The value attribute closes early and onfocus/autofocus become real attributes - a working injection needing no script tag. After the fix the same input emits

  value='&#039; onfocus=alert(document.domain) autofocus x=&#039;'

which is inert text inside the attribute.

All four are escaped with htmlspecialchars(..., ENT_QUOTES, 'UTF-8'). ENT_QUOTES matters because three of the four attributes are single-quoted, and the default flags in PHP 8.1+ would cover it but stating it keeps the intent explicit. sort, menu and offset sit inside heredocs, which cannot call functions, so those three are escaped into locals just above their heredoc.

Checked and deliberately not changed:

  $sort also reaches SQL, but apply_sort_sqlOverride() already whitelists the
  direction to asc/desc and requires an unknown field to match a strict
  identifier regex, returning '1' otherwise. Not an injection, so left alone.

  $column[align], $column[headerName] and $column[description] are also
  interpolated unescaped, but they come from the column configuration in each
  caller rather than from the request. Out of scope for this change; worth its own
  look if any caller ever derives a column definition from user input.

  The stray commas in name='sort[{$id}]', value=... at :966-967 are pre-existing
  and untouched.

Tests: tests/characterization/includes/GridSearchTermEscapingTest.test.php - 9 cases covering the quote-breaking payload that was live, tag injection, a double-quote variant, an ampersand, and four benign terms that must still round-trip so the search box keeps showing what the user typed.

Confirmed the test fails against unpatched grid.php - 5 of 9 cases - rather than only passing against the fix.

Suite: 303 tests, 1 error, 7 failures - the same pre-existing NewlabelCharacterizationTest error and UsdecimalCharacterizationTest failures master produces on its own. php -l clean; the two "optional parameter before required" deprecations at :696 and :824 are pre-existing and unchanged, verified by linting master's copy.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved protection for table and data-grid controls by safely escaping request-based values displayed in search fields and hidden form inputs.
  - Prevented specially crafted search terms, sorting values, menu values, and pagination offsets from altering page markup or being interpreted as HTML.
  - Preserved normal handling of standard search text, including accented characters, spaces, and empty searches.

- **Tests**
  - Added coverage confirming safe rendering of special characters and common markup-injection attempts in table search fields.


ZaynSaul (changes requested):
The 4 vectors this PR targets (search term, sort, menu, offset in includes/grid.php) are correctly and completely fixed — each escaped with `htmlspecialchars($v, ENT_QUOTES, 'UTF-8')` into the right context, no double-escaping regression.

Requesting changes because the PR's own title ("escape the request-derived values grid.php writes into HTML attributes") is broader than what the diff covers, and there's a live gap inside that same stated scope: `save_column_setup()` (~L1522) stores `headerName`/`description`/`align` from `$_POST['rows'][...]` with no escaping, into `datatables.column_setup`. On the next render, `render_table_headers()` echoes them unescaped (L1062/1065/1068/1086, including a CSS/attribute-breakout context on `align`).

PoC: POST `rows[<id>][1][headerName]=<script>alert(document.cookie)</script>` to the grid's `kolonner` menu → stored per-user → executes on every later view. Since this codebase has zero CSRF tokens anywhere (grepped repo-wide), a victim can be made to submit this form without realizing it — this is attacker-controlled stored XSS against another logged-in user, not just self-XSS. The PR body's claim that these fields "come from the column configuration... rather than the request" is incorrect for this persisted per-user path.

What's needed: apply the same escaping this PR already uses to `save_column_setup`/`render_columns`/`render_table_headers`'s column-config echoes.

Separately, noting for the record (not this PR's responsibility to fix): the repo-wide absence of CSRF tokens is what turns this into an attacker-exploitable bug rather than self-XSS — worth its own ticket/sign-off, independent of whether this specific gap gets fixed here.

**Decision:** Request Changes.
**Blocker:** `save_column_setup()`/`render_table_headers()` column-config fields (`headerName`/`description`/`align`) unescaped — stored XSS, CSRF-reachable. Owner: author.
**Testing:**
- Verified by me: all 4 targeted vectors' escaping correctness; the column-setup stored-XSS chain end-to-end (save → DB → render); repo-wide absence of CSRF tokens (grep).
- Author-reported, not independently reproduced: the 9/9 passing characterization tests, "5 of 9 fail on unpatched" claim.
- Unverified by anyone: live/browser behavior of either the fix or the gap found here.

**Nonblocking follow-up:** the repo-wide CSRF-token gap deserves its own ticket/security review — separate ask from this PR's fix, not blocking it.

ZaynSaul:
@momocoder14 bumping this - the save_column_setup()/render_table_headers() stored-XSS gap from the review is still there (headerName/description/align saved unescaped and echoed unescaped on the next render). The 4 vectors this PR targets are fine, just needs the same escaping applied to those three fields before this can merge.

